PR #37, opened 2021-05-26
Implement scheduling API

galleon:
Code for #24 implementing scheduling API and fixing examples


galleon:
- Changes for including scheduling API

PR #48, opened 2021-08-25
Change build worflow

galleon:
Remove all upload tasks so that the wf can be run by users not having write access (which is needed for PRs)


dbarbier:
Not related to this PR, but maybe we could discuss which checks are useful for PRs? IMO all 6 tests and build-docs should be added too.

PR #49, opened 2021-08-26
autodoc: fix notebooks download links + other improvements

nhuet:
- fix  notebook download links
- notebooks generation: add a title cell and allow magic commands
  ex:
  ```
    # %%
    # %matplotlib
  ```
  transforms into a code cell with magic command `%matplotlib` 
- do not version files autogenerated by autodoc.py
- create subdirectories like docs/reference when missing (e.g. at first doc build)
- the script can be now launched also from main directory


neo-alex (approved):
Seems all good to me, thanks!

PR #51, opened 2021-08-26
Adding release workflow 

galleon:
The workflow is triggered on v tags

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -262,7 +262,7 @@ jobs:
       - name: Publish distribution 📦 to PyPI
         env:
           PYPI_TOKEN: ${{ secrets.PYPI_PASSWORD }}
-        if: env.GITHUB_REPOSITORY  == 'airbus/scikit-decide' && env.PYPI_TOKEN != ''
+        if: github.repository == 'airbus/scikit-decide' && env.PYPI_TOKEN != ''
         uses: pypa/gh-action-pypi-publish@master
         with:
           password: ${{ secrets.PYPI_PASSWORD }}


.github/workflows/release.yml
```diff
@@ -0,0 +1,320 @@
...
+      - name: generate documentation
+        run: yarn global add vuepress && yarn install && yarn docs:build && touch docs/.vuepress/dist/.nojekyll
+
+      - name: Deploy 🚀
+        uses: JamesIves/github-pages-deploy-action@releases/v3
+        with:
+          ACCESS_TOKEN: ${{ secrets.gh_access_token }}
+          BRANCH: gh-pages # The branch the action should deploy to.
+          FOLDER: docs/.vuepress/dist # The folder the action should deploy.
+          GIT_CONFIG_EMAIL: guillaume.alleon@gmail.com
```
fteicht, line 317: @galleon To change to an Airbus email address as soon as we have one.

.github/workflows/release.yml
```diff
@@ -0,0 +1,320 @@
...
+
+    steps:
+      - name: Download artifact
+        uses: actions/download-artifact@v1.0.0
+        with:
+          name: wheels
+
+      - name: Publish distribution 📦 to PyPI
+        env:
+          PYPI_TOKEN: ${{ secrets.PYPI_PASSWORD }}
```
fteicht, line 264: @galleon Which account does it use by default (say when the script is run from Airbus' scikit-decide repository)?

.github/workflows/release.yml
galleon: Yes, you can always open an issue as a reminder.

.github/workflows/release.yml
galleon: It uses `scikit-decide` project on PyPI. The secret can be changed from github admin page. For the time being we are both users on this `scikit-decide` project.

fteicht (approved):
It's good for me, thanks for your work @galleon !

PR #52, opened 2021-08-30
Fix builds in editable mode

dbarbier:
README instructions tell to compile scikit-decide by running

    poetry install --no-root --extras all
    poetry build

But as a developer this is frustrating because we have to wait for a long build time when testing changes in source code. We want instead to build scikit-decide in editable mode (like `pip install -e`) by running instead

    poetry install --extras all

This PR makes this possible.


galleon (approved):
Looks good to me

PR #53, opened 2021-09-02
Run various linters on pull requests

dbarbier:
This PR is quite self-explanatory, it adds a new "Code style" check at the bottom of this page in the checks tab.

Mark this PR as a draft for now, we have to discuss which linters do we want to apply and on which files, eg. do we want to reformat documentation and notebooks too?  Also this PR should be applied only when there is no conflict with other PRs.

Checks will be performed on PRs created after this PR is merged, and if some check fails, author will have to fix source files.


dbarbier:
It is fairly easy to reformat C++ source files as well with clang-format, but we need to decide on some coding style guidelines.  A comparison between most used guidelines can be found at https://github.com/motine/cppstylelineup, please tell which one yu want.

galleon:
@fteicht or @jeromerobert can you comment and contribute on which c++ style to apply on C++ files

fteicht:
Thanks @dbarbier for digging int the topic. Personally I am used to the LLVM style that I like a lot. The Google style is fine to me too. What do you think @jeromerobert ?

jeromerobert:
LLVM is also my prefered one. I used Webkit for a while but 2 spaces indentation is definitely better. Google is good except when it create one line `if` statments. Chromium is also good but has a bit too many (IMHO useless) line breaks.

dbarbier:
Okay I just added some commits to call clang-format with LLVM style on all C++ files in cpp/src directory.  As mentioned in d636152 I had to disable sort of includes, otherwise there are build failures.

dbarbier:
Force-pushed to fix conflicts

dbarbier:
Black is now run on docs and examples too, and data files are no more processed.  This PR is no more a draft and can be reviewed.

fteicht:
@dbarbier I don't see any changes in the C++ files so I guess I have to run some script manually on the source code on my side? Why the linted C++ files are not pushed in the PR?

galleon:
Might also be interesting to have a pre-commit hook for `json` as you suggested in #72. I would suggest to have 4 spaces indentation (default in VSCode).

dbarbier:
It is IMO more important to have this PR merged and source files modified without messing with other PRs.

PR #54, opened 2021-09-07
Gym notebook

nhuet:
Notebook presenting how to define a gym environment based domain and solve it with scikit-decide.


nhuet:
Still missing some explanation about  algorithms used to solve the problem and about the problem itself. Help for that would be greatly appreciated !

nhuet:
I introduced IW solver as suggested by  @fteicht , but it seems not working even though i tried to reproduced what was done in full_multisolve.py

nhuet:
This notebook needs ffmpeg to work as an additional dependency. It already exists on colab and i added it on binder but i can only state at the beginning of the notebook that local jupyter users need to install it as it is platform dependent.

nhuet:
Here are the link to test it on colab and binder:
- binder: [Binder](https://mybinder.org/v2/gh/nhuet/scikit-decide-nb-with-binder/master?urlpath=git-pull%3Frepo%3Dhttps%253A%252F%252Fgithub.com%252Fnhuet%252Fscikit-decide%26urlpath%3Dtree%252Fscikit-decide%252Fnotebooks%252Fgym_demo.ipynb%26branch%3Dgym_nb)
- colab: [Open In Colab

galleon:
@nhuet binder seems to be super slow. I have tried against your link above, il has run for 5mn closing with this error:
```
Found built image, launching...
Launching server...
Failed to connect to event stream
```
Any idea what could be the problem ?

fteicht:
@nhuet can you briefly describe the issue with IW?

nhuet:
> @nhuet binder seems to be super slow. I have tried against your link above, il has run for 5mn closing with this error:
> 
> ```
> Found built image, launching...
> Launching server...
> Failed to connect to event stream
> ```
> 
> Any idea what could be the problem ?

It can happen sometimes, have you  tried to relaunch?

nhuet:
@neo-alex and @fteicht : i inserted your explanations texts. So the notebook is ready for a first review. Some texts are probably still missing (i put some placeholder to show where)

nhuet:
@neo-alex : more importantly, i included a text telling that PPO is doing well in this context but actually with its current settings, this is the only solver that does not work ! (even though it is a RL one) i have total_timesteps=50000, what else can i tune to make it work? (all of this is random and i did not fix the seed for now, but PPO was never working during my tests,while  CGP and IW were always working)

nhuet:
NB: using our own rollout avoid having to use monitor and ffmpeg, and fix display issue with IW. Actually we are back to a "real time " rendering, no more movie prepared before seeing something. But this is not so smooth on binder (compared to local jupyter session). Still working though

nhuet:
> * CGP is not working better than PPO on my binder instance! So big warning here. It seems that sometimes it does not work (how much time? Were I unlucky? But if it happens, we have to warn the reader at least.)

Indeed! Also for me with binder oops. By doubling the n_it it seems to work though and i added a warning as suggested. (on my laptop it never failed me, strange...)

nhuet:
Actually, it depends on episodes. So perhaps it is better to launch several episodes, instead of increasing the number of iterations during solve.

PR #55, opened 2021-09-14
Update and synchronize README.md and online doc about installation steps

nhuet:
Online doc and README.md diverged about how to install the library and are not up-to-date.


README.md
```diff
@@ -32,29 +32,65 @@ Scikit-decide is an AI framework for Reinforcement Learning, Automated Planning
...
-On MacOS/Linux:
-
-    python3 -m venv /path/to/new/virtual/environment
-    source /path/to/new/virtual/environment/bin/activate
-
-On Windows:
-
-    py -m venv \path\to\new\virtual\environment
-    \path\to\new\virtual\environment\Scripts\activate
+    conda create --name skdecide python=3.7
```
galleon, line 35: Why moving from pyenv to conda? 
Do we need `conda`at all ?

nhuet: We do not *need* it. But 
- it was the solution proposed in the online doc => it can be changed also in online doc, no pb
- the former solution in readme was actually not selecting a version of python 
- it was perhaps easier than  
  - to install pyenv and choose an appropriate version of python
  - and to use a second tool  to create a virtual environment
- Other possibilities pipenv or poetry are more designed for developping a project (like we do for developper mode installation)

Notice that we *do* use pyenv for local install in developper mode, but it was the only solution that worked in the end for me so it seemed important to share it as is.

However, i also think it could be nice to have some steps shared between the 2 modes (recommened install vs developper install), so perhaps pyenv + venv would be better in the end (but in this case we need to change online doc also). I let you decide :)

POPOGO:
I'd be interested to know how to test rapidly local changes in my source code like it is possible when installing a library using "pip install --editable", because i think when using poetry you install the wheel but if you modify the python source code you have to recreate a whl (so takes 30 minutes)

dbarbier:
> I'd be interested to know how to test rapidly local changes in my source code like it is possible when installing a library using "pip install --editable", because i think when using poetry you install the wheel but if you modify the python source code you have to recreate a whl (so takes 30 minutes)

Follow instructions in README.md and that should be just fine

nhuet:
> I'd be interested to know how to test rapidly local changes in my source code like it is possible when installing a library using "pip install --editable", because i think when using poetry you install the wheel but if you modify the python source code you have to recreate a whl (so takes 30 minutes)

This is more or less what i explained in developper mode on README.md. I added more details so that it should work now. And i remove the option "--no-root" from poetry install which actually prevented to install the scikit-decide code itself (in editable mode). 
This section does not appear in the online doc for now though. We have to decide wether we want to have it or not on the online doc. As this is relevant only for contributors to the library and not for normal users of the library (whom we want to make use the official release on pypi)

README.md
galleon: Ok @nhuet Let's follow your idea and share what can be shared. So remove `conda` and replace by `pyenv`. We need also to change the online doc.

nhuet: I just realized the following issue: pyenv is not for windows users. It seems that a corresponding tool [pyenv-win](https://github.com/pyenv-win/pyenv-win) has been developed for windows. I need to check on my windows laptop that commands still are working as expected though.

nhuet:
I updated installation notes by using pyenv instead of conda as suggested by @galleon, and synchronized docs/guide/README.md install section with root README.md. I did not mention developer mode in online doc as it seems that this doc was more intended for users (either end-users or domain/solver developers) than for core library contributors.

Remaining tasks:

- [ ] check how it works on windows 
- [ ] see if warnings at the end of playground section are still needed.
  > **Warning**: some domains/solvers might require extra manual setup steps to work at 100%. In the future, each scikit-decide 
  > hub entry might have a dedicated help page to list them, but in the meantime please refer to this:
  > 
  > - [domain] OpenAI Gym ones -> [gym](http://gym.openai.com/docs/#installation) for loading Gym environments not included by default
  > - [solver] PPO: Proximal Policy Optimization -> see [Stable Baselines installation](https://stable-baselines.readthedocs.io/en/master/guide/install.html)
  > - [solver] IW: Iterated Width search (same for AOstar, Astar, BFWS) -> special C++ compilation (see Installation 2.c. above)
  >

nhuet:
I tried quite a bit to make it work on windows:
- pyenv-win can replace pyenv, but i was compelled to use UI to set the environment variables i will not describe in details the process in the readme
- i had less success with poetry which kept arising errors (perhaps that could be removed but i think the whole poetry thing took already too much time for now)
=> I think i will skip the windows process for developer mode as it concens only core developers (so not so much people) that would be interested in doing that on windows platform (even less to my mind).
=> I will instead put a disclaimer in section developer mode explaining it has not been tested on windows for now (but surely windows-fluent users should be able to make it work)

nhuet:
Because i am not aware of all plugins :)
It was working like i wrote, so i did not ask for more. 
I see that it is a plugin existing for pyenv to manage virtualenv directly via pyenv. Though i am not sure it is working on pyenv-win. And at least my command (python -m venv) will work even for people that did not choose to use pyenv to select their python version...

I'd like to have other people opinion on that.

nhuet:
I let only the warning for gym domain at the end of playground section.

I think i am more or less done with that PR (except potentially for the virtualenv plugin for pyenv).
I would like to share the generated online guide with you for review but for now pushing on master does not deploy the docs anymore (since it is done during release). So you will have to build docs locally or wait for the nightly deployment of docs to work ;)

POPOGO:
Ok I checked each commit and their content, the commit message are coherent and clear.
 I want to test the developer mode on my side today.

PR #56, opened 2021-09-17
upgrade environment for maze and gym notebooks

nhuet:


nhuet:
Actually i used these dependencies in my version of notebooks and i do not see why it would be so bad to "risk" to have a bit too many dependencies? What is the risk ? At least we do not need to rebuild it once it is done. 
For matplotlib, i put it explicitely because i did not realize it was part of the dependencies (probably because of the "official" maze domain in the hub) which does not change anything as you say. I can remove that if you prefer even though i did not think it was worth it...

galleon:
> Actually i used these dependencies in my version of notebooks and i do not see why it would be so bad to "risk" to have a bit too many dependencies? What is the risk ? At least we do not need to rebuild it once it is done.
> For matplotlib, i put it explicitely because i did not realize it was part of the dependencies (probably because of the "official" maze domain in the hub) which does not change anything as you say. I can remove that if you prefer even though i did not think it was worth it...

Je pense que c'est une bonne pratique de mettre uniquement les paquets utiles. D'autant plus sur des notebooks qui pourraient être utilisés par des utilisateurs sous divers environnements (que nous n'avons pas testé). Je préfère les ajouter au fil de l'eau - cela ne rend que 15 mn à construire.

dbarbier:
@nhuet Do not lose your time on this issue, if your notebooks need a different environment, use a branch on your fork.

PR #65, opened 2021-09-30
Maze notebook: educational version

nhuet:


dbarbier:
@nhuet Can you please provide a link to binder?

nhuet:
[Binder](https://mybinder.org/v2/gh/nhuet/scikit-decide/binder_env?urlpath=git-pull%3Frepo%3Dhttps%253A%252F%252Fgithub.com%252Fnhuet%252Fscikit-decide%26urlpath%3Dtree%252Fscikit-decide%252Fnotebooks%252Fmaze_tuto.ipynb%26branch%3Dmaze_nb_seq)

notebooks/maze_utils.py
```diff
@@ -0,0 +1,164 @@
...
+            fig.set_figheight(600)
+        if image is None:
+            image = ax.imshow(image_data)
+        else:
+            image.set_data(image_data)
+            image.figure.canvas.draw()
+        return ax, image
+
+    def is_an_empty_cell(self, i, j):
+        return self.maze_array[i][j] == Maze._empty_cell
```
dbarbier, line 164: should be `self.maze_array[i, j]`

galleon:
Here is the link using the local (on this repo) binder env.

[Binder](https://mybinder.org/v2/gh/airbus/scikit-decide/binder?urlpath=git-pull%3Frepo%3Dhttps%253A%252F%252Fgithub.com%252Fnhuet%252Fscikit-decide%26urlpath%3Dtree%252Fscikit-decide%252Fnotebooks%252Fmaze_tuto.ipynb%26branch%3Dmaze_nb_seq)

notebooks/maze_tuto.ipynb
```diff
@@ -0,0 +1,642 @@
...
+   "source": [
+    "import io\n",
+    "from enum import Enum\n",
+    "from typing import NamedTuple, Optional, Any, List\n",
+    "from time import sleep\n",
+    "from collections import deque\n",
+    "import random\n",
+    "from math import sqrt, ceil\n",
+    "\n",
+    "from IPython.display import Markdown, display, clear_output\n",
```
galleon, line 37: Sanitize imports - Markdown, plt ... are not used. To name a few.

notebooks/maze_tuto.ipynb
```diff
@@ -0,0 +1,642 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# size of maze\n",
+    "width = 25\n",
+    "height = 19\n",
+    "# generate the maze\n",
+    "maze = Maze.generate_random_maze(width=width, height=height)\n",
+    "# starting position\n",
+    "i_start, j_start = 1, 1\n",
```
galleon, line 87: Might make sense to store start and final positions as tuples. This would be consistent with State definition

notebooks/maze_tuto.ipynb
```diff
@@ -0,0 +1,642 @@
...
+   "outputs": [],
+   "source": [
+    "def my_rollout(domain, solver, max_steps, pause_between_steps=0.01):\n",
+    "    # Initialize episode\n",
+    "    solver.reset()\n",
+    "    observation = domain.reset()\n",
+    "\n",
+    "    # Initialize image\n",
+    "    figure = domain.render(observation)\n",
+    "    display(figure)\n",
```
galleon, line 524: Use display(domain.render(observation)) as earlier - same later.

notebooks/maze_tuto.ipynb
```diff
@@ -0,0 +1,642 @@
...
+    "    max_steps = maze.width * maze.height\n",
+    "    print(f\"** Playing solution  with max_steps={max_steps} ...\", flush=True)\n",
+    "    rollout(\n",
+    "        domain,\n",
+    "        solver,\n",
+    "        num_episodes=1,\n",
+    "        max_steps=max_steps,\n",
+    "        max_framerate=None,\n",
+    "        outcome_formatter=None,\n",
+    "        render=False,\n",
```
galleon, line 430: I did change this to True ... (assuming a user would try to do so) but the graphic is not displayed.

galleon: I think it would be best to solve and then write our own roll out ...

notebooks/maze_tuto.ipynb
nhuet: Yes but, maze_utils.py is not aware about State definition. Actually this notebook is somehow about wrapping an existing class (Maze) that do no know about scikit-decide, in a scikit-decide domain. that's why Maze.render cannot use State as input. I can still use tuples though, i just need to be cautious to avoid confusion between the tuple and State class...

notebooks/maze_tuto.ipynb
nhuet: I did that, (writting the rollout) but in A* (following the narratives). I can put it here if necessary. 
The bad part is that we need to do that at the same time as the solve because of this "with ..." syntax that scikit-decide want to enforce when using solvers.

notebooks/maze_tuto.ipynb
dbarbier: I just pushed a commit to #53 to sanitize notebooks too.

notebooks/maze_tuto.ipynb
dbarbier: The basic usage is to call from from inside a context manager; it is okay IMO to emulate it without a context manager, but the important point is to call `solver.close()` at the end.

notebooks/maze_tuto.ipynb
```diff
@@ -0,0 +1,699 @@
...
+    "        )\n",
+    "        return self._image.figure\n",
+    "\n",
+    "    def heuristic(self, s: State) -> Value[float]:\n",
+    "        \"\"\"Heuristic to be used by search algorithms.\n",
+    "\n",
+    "        Here Manhattan distance to goal.\n",
+    "\n",
+    "        \"\"\"\n",
+    "        return Value(cost=sqrt((self.end.x - s.x) ** 2 + (self.end.y - s.y) ** 2))"
```
dbarbier, line 311: This is usual Euclidean distance, not Manhattan.  BTW do we really need sqrt?

nhuet: Euh you are right of course. I juste copy/paste a comment from Alex without thinking further... my bad

nhuet:
@neo-alex @fteicht : i think i took all commentary made into account, so it should be now your turn to review.
NB:
- missing explanation texts are shown via placeholders like that: **_text needed_** (bold + italic)
- the syntax using `with` for the solver prevent splitting solver part from rollout part -> do you think it would be a good idea to split it, even though not using your preferred syntax? (you will also see that my explanation about that syntax is quite clumsy)
- we should also think to fix an appropriate total_timesteps for ppo so that it is still clear that is not about to work, and fast to compute at the same time

nhuet:
Finally I decided to simplify explanations by separating solve and rollout in 2 different cells. Still i added a comment at the end to explain how to automatically call cleanup thanks to `with`. You will see how this last commit makes the notebook much more readable (imho).

fteicht (changes requested):
Thank you for this amazing notebook!
Here are a few minor comments:

- remove the '.' at the end of the first bullet sentence in the introduction text
- change "About maze problem" for "About the maze problem"
- change "bottom_right" for "bottom-right" in the paragraph titled "About maze problem"
- change "recognized as scikit-decide domain" for "recognized as a scikit-decide domain"
- change "We also specify type of states [...]" for "We also specify the types of states [...]"
- should we explain why this is a `DeterministicPlanningDomain`? (deterministic starting states and transitions, white box transition model, definition of goal states)
- "Perhaps more details about why this intermediate class D is needed (autocast?)" => yes, I believe this is definitely required, I let @neo-alex comment on this.
- "Text needed to explain why methods to overload have leading underscores" => I guess @nhuet will do it otherwise please ask @neo-alex to comment
- "we will need a domain factory recreating the domain at will": we can mention that it is for instance useful for parallel solvers which create identical domains on separate processes by using the domain factory
- change "allowing to use" for "allowing us to use"
- change "that is make use of" for "that makes use of"
- change "which give access to" for either "which gives access to" or "giving access to"
- change "Here we choose Proximal [...]" for "Here we choose the Proximal [...]"
- "Explanation needed about the not intuitive syntax to use here (needed for autocast?)" => ask @neo-alex to explain
- change "Rolling out a solution" for "Rolling out the solution (found by PPO)"
- change "see utils.py module" for "see the utils.py module"
- "This is probably because it is not using the domain characteristics to their finer level.": this is actually to the fact that he reward is sparse (you get rewarded only when you reach the goal) and this is nearly impossible for this kind of RL algorithm to reach the goal just by chance without shaping the reward.
- "well known to be suited to this kind of problem": because it exploits the knowledge of the goal and of heuristic metrics to reach the goal (e.g. euclidean or Manhattan distance)
- "previously defined heuristic": I can't see where it was previously defined or presented
- change "Training solver on domain" for "Training the solver on the domain"
- change "Rolling out a solution" for "Rolling out the solution (found by Astar)"
- change "more infos" for "more information"
- change "all reward" for "all rewards"
- In the list of reasons why Astar performs well on the maze, we should mention that it exploits the knowledge of the existence of the goal (aka the space reward in this domain)
- change "from RL community" for "from the RL community"
- change "That's why" for "That is why"
- "define the domain with the finer granularity possible": and also to use the solvers that can exploit at most the known characteristics of the domain

fteicht:
After reading the conclusion of this notebook, I am thinking that we should maybe create another notebook in the future to highlight a problem where RL is especially better than other methods. Because at the moment we only have tutorial notebooks which seem to constantly demonstrate the opposite. On pure control problems with continuous rewards (e.g. CartPole), RL is typically better, and this not the only case. It's obviously true for problems for which we don't know the model - but in that case they won't be other options neither.

nhuet:
I made all changes requested by @fteicht . It remains now only two explanations let to @neo-alex :
- details about why this intermediate class D is needed (autocast?)
- explanation about solve_with syntax : why this is called from domain class and not instance or even solver ? 
  I think this is also for the autocast feature, but not sure how to explain it properly. 
  Indeed for a new user, a more intuitive syntax would be `solver.solve(domain_factory=lambda: domain)` as @POPOGO pointed out to me.

dbarbier:
[...]
>     * explanation about solve_with syntax : why this is called from domain class and not instance or even solver ?
>       I think this is also for the autocast feature, but not sure how to explain it properly.
>       Indeed for a new user, a more intuitive syntax would be `solver.solve(domain_factory=lambda: domain)` as @POPOGO pointed out to me.

This is explained below in the "Cleaning up  the solver" section, doesn't it?

nhuet:
> This is explained below in the "Cleaning up the solver" section, doesn't it?

Nope. What you are talking about is the syntax involving "with ...". What I was talking about is why using
```python
MazeDomain.solve_with(solver, domain_factory)
```
instead of 
```python
solver.solve(domain_factory)
```
The latter being more intuitive to my mind. Especially why the need to call the domain class and not the instance itself.
Actually @POPOGO even told me he was using `solver.solve(domain_factory)` syntax but it is not how it should be done if i remember. And i would like to explain exactly why here. (and thus i need @neo-alex help to understand it perfectly)

notebooks/1_maze_tuto.ipynb
```diff
@@ -0,0 +1,740 @@
...
+    "*NB: To know the methods not yet implemented, one can either use an IDE which can find them automatically or the [code generators](https://airbus.github.io/scikit-decide/guide/codegen.html) page in the online documentation, which generates the corresponding boilerplate code.*"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class MazeDomain(DeterministicPlanningDomain, UnrestrictedActions, Renderable):\n",
```
neo-alex, line 190: Should be `class MazeDomain(D)` instead

neo-alex (changes requested):
Just one mistake to correct please (see comment) and linting to apply, then OK for merge!

notebooks/1_maze_tuto.ipynb
nhuet: Oops, again a test that remained behind, my bad... Changed!

PR #70, opened 2021-10-07
Implement multi docs

galleon:
- [x] Add workflow to generate doc
- [x] Add new dropdown widget in navbar to select doc version

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -227,6 +227,7 @@ jobs:
 
   build-docs:
     needs: [test-unix, test-macos, test-windows]
+    if: ${{ github.ref == 'refs/heads/master' }}
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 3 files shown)


dbarbier:
Why is a new workflow needed?  We already have build.yml and release.yml.  BTW release.yml should have been modified to deploy.

galleon:
> There are at least 2 ways to build versioned documentation:
> 
> * Static: available versions are hardcoded when building docs, which means that an older documentation can provide links to master and previous releases, but not future ones
> * Dynamic: all available versions are retrieved by clients.  This comes at a cost, Github limits the number of API requests, so the dropdown menu may be unavailable when displaying lots of pages.
> 
> Your commit message for [5ab99ab](https://github.com/airbus/scikit-decide/commit/5ab99ab4014a5ea39c92d7100a13db22674924b0) is way too terse. You should describe the different options, and why you implemented it that way.

Will improve commit description.

galleon:
> Why is a new workflow needed? We already have build.yml and release.yml. BTW release.yml should have been modified to not deploy.

`docs.yml` has been removed. Docs is generated through `build.yml`when on master or through `release.yml` workflows.

Issue #68, opened 2021-10-05
collect_env.py script missing but required by the bug report template


fteicht:
## 🐛 Bug

GitHub's bu report template asks for pasting and copying the output of the script `https://raw.githubusercontent.com/airbus/scikit-decide/master//utils/collect_env.py` which does not exist.

## To Reproduce

Create a new issue and navigate to the end of the template proposed by GitHub, you will see the link to the missing file.

## Expected behavior

The script `collect_env.py` under a folder `utils` (which does not exist either) should exist.

PR #73, opened 2021-10-11
Adding collect_env.py (modified from PyTorch)

galleon:
Solving issue #68

### DIFF
--- a/utils/collect_env.py
+++ b/utils/collect_env.py
@@ -1,0 +1,469 @@
+from __future__ import print_function
+
+# Unlike the rest of the PyTorch this file must be python2 compliant.
+# This script outputs relevant system environment info
+# Run it with `python collect_env.py`.
+import datetime
+import locale
+import re
+import subprocess
+import sys
+import os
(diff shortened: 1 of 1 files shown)


galleon:
It was proved useful in the past to capture the main packages users have in their environment when using `scikit-decide`.  They sometimes update some packages which break our library. It avoid sometimes tedious exchange to capture this info. The script captures at the moment only basic info + info on pytorch and stable-baseline but it cam be improved as we are facing new issues. Also useful, when people are rebuilding from sources where compilers can play a role.

fteicht:
This script does not add much work on the bug reporter shoulders, yet it can sometimes bring valuable information to us regarding possible causes. Therefore I see more added value than the inverse.

fteicht:
Also, scikit-decide embedding C++ code, it is important to check that the Python environment of the user is compatible with the compiler used to compile the C++ extension of the library. It is the case for the vast majority of the cases, but we never know (and it has happened in the past).

PR #72, opened 2021-10-11
gal/multi docs

galleon:
Follow up of #70 that cannot be re-opened :-(

- This PR generates a static web site for the `scikit-decide` documentation.

- The theme is modified to include a dropbox for version number selection.

- Admissible versions number are retrieved through a WS call to GH.

PR #75, opened 2021-10-12
[Trivial] Make doc generation easier for a VuePress developper

galleon:
A VuePress developper will use either `yarn run docs:dev` to test the documentation or `yarn run docs:build`to generate it. It seems natural to integrate the generation of the `_state.json`  file read by `docs/.vuepress/enhanceApp.js` into that process.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -252,9 +252,5 @@ jobs:
         run: |
           pip install --pre --find-links ./wheels/ "scikit-decide[all]"
 
-      - name: Generate Python docs
-        run: |
-          (cd docs; python autodoc.py)
-
       - name: generate documentation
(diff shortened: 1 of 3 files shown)


neo-alex:
> It seems natural to integrate the generation of the `_state.json`

I see your point @galleon but one could argue that all files generated by `autodoc.py` (incl. `_state.json`) are necessary to have the full documentation properly running, thus I wonder if it is worth doing an exception for this file?

dbarbier:
That being said, the same problem occurs if one runs `vuepress dev docs` directly.

galleon:
Included in #72 as suggested

dbarbier:
That is not what I suggested.

dbarbier (approved):
Thanks, this PR is now minimal and should hopefully not conflict with other PR, it can be merged.

PR #80, opened 2021-10-19
[trivial] Treat version numbers as strings

dbarbier:
Without quotes, version number is considered as a float, which will cause trouble with Python 3.10.

See https://github.com/github/docs/issues/11019


galleon (approved):
Looks Good To Me

PR #85, opened 2021-10-25
Integration of new notebooks in doc

nhuet:
The new notebooks available in "notebooks/" folder will replace the previous small notebooks generated from python scripts in "examples/" folder.

The autodoc.py script is used to:
- list the notebooks and sort them alphabetically
- extract from first cell (which should be a markdown cell)
  - a title (first line if starting with "# ")
  - a description (the other lines)
- generate links
  - to get the notebook from github
  - to launch the notebook with binder
- modify the "guide" page by inserting this list with titles, description, and links,
  in "examples" section (a template will now be used to generate the docs/guide/README.md)
  More precisely, docs/guide/README.md is generated from docs/guide/README.md.template
  where placeholder [[notebooks-list]] is replaced by this list of notebooks

In order to have working links, some environment variables are needed to specify
- repository and branch used to host the binder environment
  -> if not defined, autodoc.py falls back to airbus/scikit-decide:binder
- repository and branch used to fetch the notebooks contents
  -> if not defined, autodoc.py does not generate github+binder links
     and issues a warning

README.md is updated to:
- explain how to define properly the environment variables to build the
  doc
- mention the notebooks in examples section

To test this PR, you can build the documentation locally by typing from root directory:
```shell
export AUTODOC_BINDER_ENV_GH_REPO_NAME="airbus/scikit-decide"
export AUTODOC_BINDER_ENV_GH_BRANCH="binder"
export AUTODOC_NOTEBOOKS_REPO_URL="https://github.com/nhuet/scikit-decide"
export AUTODOC_NOTEBOOKS_BRANCH="nh/doc_notebooks_integration"
yarn install
poetry run python docs/autodoc.py
yarn docs:dev
```
provided you have already installed the library in developer mode with poetry as explained in README.md.

To generate the online doc, the environment variables "AUTODOC_NOTEBOOKS_REPO_URL" and "AUTODOC_NOTEBOOKS_BRANCH" will have to be properly defined by the workflow to correspond to the repository and branch from which it is triggered, and for which the doc is generated. This update of online doc generation should be done in another PR.

PR #86, opened 2021-10-25
[Trivial] Update gym notebook for online doc integration

nhuet:
Slight modifications to gym notebook for good integration in online doc thanks to PR #85:
- The name is prefixed by 2_ to be seen as second notebook
- The description is slightly modified (the order of sentences changed,  "three reasons" changed to "two reasons" as there are only 2 stated


neo-alex:
Approved for merge (after linting)!

PR #90, opened 2021-10-28
[trivial] Prevent build.yml workflow to trigger on tags push

nhuet:
As this trigger is already handled by release.yml for release tags.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,8 @@ name: 🔨 Build scikit-decide
 
 on:
   push:
+    branches:
+      - "*"
   pull_request:
   workflow_dispatch:
 


nhuet:
I switched to tags-ignore = "*". Both solutions (the latter and branchs = "*") worked but the new version (pointed out to me by @dbarbier) is more explicit.

nhuet:
Actually i will revert to previous version as 
```yaml
on:
  push:
    tags-ignore: 
      - "*"
```
actually prevent triggering on tags but also on branches (this is odd i know), though still working for PR
whereas
```yaml
on:
  push:
    branches: 
      - "*"
```
work as we want: triggering on all branches push but not on tags push (but still on PR).

nhuet:
I do not know why but my branch `nh/prevent-build.yml-for-tags` actually does not trigger any action on my repo whereas:
- it works for PR as we can see here
- another branch "test" at the same commit does trigger build.yml (as it was expected)
- i suspected a problem with the branch name with the dot in it, but yet another branch with a simpler name `nh/prevent-build-for-tags` pointing also at the same commit does not trigger build.yml either.

Very strange... :thinking:

PR #93, opened 2021-10-29
Add benchmarking tutorial notebook

neo-alex:


galleon:
If we want this tuto to run on `binder` we might have to add a better reporting for failing algorithms and better control the limited computing resources.

galleon:
Superseded by #125

PR #95, opened 2021-11-02
Nightly wheels

galleon:
## 🚀  Motivation

As time between release can be a couple of months, it is important to publish wheels for the different supported infrastructures on a shorter term basis

## 🛠️ What does this PR do?

- Create a `nightly` release once a day each time `master` has changed 
- Publish a zip containing all wheels corresponding to the change
- Once a new named release is created, the nightly ones are deleted

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,8 @@ on:
       - "**"
   pull_request:
   workflow_dispatch:
+  schedule:
+    - cron:  '45 1 * * *'
 
 jobs:
 
@@ -298,7 +300,7 @@ jobs:
           echo "AUTODOC_BINDER_ENV_GH_BRANCH=${AUTODOC_BINDER_ENV_GH_BRANCH}" >> $GITHUB_ENV
           echo "AUTODOC_NOTEBOOKS_REPO_URL=${AUTODOC_NOTEBOOKS_REPO_URL}" >> $GITHUB_ENV
           echo "AUTODOC_NOTEBOOKS_BRANCH=${AUTODOC_NOTEBOOKS_BRANCH}" >> $GITHUB_ENV
-          # check computed variables
+          # check computed variables
           echo "Binder env: ${AUTODOC_BINDER_ENV_GH_REPO_NAME}/${AUTODOC_BINDER_ENV_GH_BRANCH}"
           echo "Notebooks source: ${AUTODOC_NOTEBOOKS_REPO_URL}/tree/${AUTODOC_NOTEBOOKS_BRANCH}"
 
@@ -343,3 +345,148 @@ jobs:
           commit-message: publish documentation
           clean-exclude: |
             "version/*"
+
+  upload-nightly:
+    if: (github.ref == 'refs/heads/master') && (github.event_name == 'schedule')
+    needs: [test-unix, test-macos, test-windows]
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/download-artifact@v2
+      with:
+        name: wheels
+        path: dist/
+
+    - run: |
+        zip -r dist.zip dist/
+
+    - uses: actions/github-script@v5
+      id: asset
+      with:
+        github-token: ${{ secrets.gh_access_token }}
+        script: |
+          const fs = require('fs');
+
+          // Get the ref for master
+          const master_sha = '${{ github.sha }}';
+          console.log(`master reference ${master_sha}`);
+
+          // Retrieve ref for tag `nightly`
+          let ref_nightly = null;
+          try {
+            ref_nightly = await github.rest.git.getRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'tags/nightly',
+            });
+
+            if (ref_nightly.data.object.sha === master_sha) {
+              return "No new nightly release";
+            }
+          } catch (err) {
+            // The tag does not exist so let's create it
+            ref_nightly = await github.rest.git.createRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'refs/tags/nightly',
+              sha: master_sha,
+            });
+          }
+
+          // Call the GitHub API to get a release by tag
+          let release = null;
+          try {
+            release = await github.rest.repos.getReleaseByTag({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              tag: 'nightly',
+            });
+            console.log(`Found release ${release.data.tag_name} ${release.data.draft} ${release.data.prerelease}`);
+          } catch (err) {
+            console.log(`Release 'nightly' not found`);
+
+            // If the release doesn't exist, create it
+            release = await github.rest.repos.createRelease({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              tag_name: 'nightly',
+              name: 'nightly',
+              body: 'Nightly release crafted with ♥️ somewhere on 🌎',
+              draft: false,
+              prerelease: true,
+            });
+
+            console.log(`Created release ${release.data.tag_name} ${release.data.draft} ${release.data.prerelease} ${doIProceed}`);
+          }
+          console.log(`Release does exist with tag ${release.data.tag_name} [${release.data.draft} ${release.data.prerelease}]`);
+
+          // At this stage both tag & release exist
+
+          // Update nightly tag
+          await github.rest.git.updateRef({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            ref: 'tags/nightly',
+            sha: master_sha,
+            force: true,
+          });
+          console.log(`Updated tag with sha ${ref_nightly.data.object.sha}`);
+
+          // Update the release
+          await github.rest.repos.updateRelease({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            release_id: release.data.id,
+            tag_name: 'nightly',
+            name: 'nightly',
+            body: 'Nightly release crafted with ♥️ somewhere on 🌎',
+            draft: false,
+            prerelease: true,
+          });
+          console.log(`Updated ${release.data.tag_name} nightly release  ${release.data.draft} ${release.data.prerelease}`);
+
+          // Get all tags and keep the newest one starting by v
+          let newest_tag = { name: 'v0.0.0' };
+
+          const tags = await github.rest.repos.listTags({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+          });
+
+          // Keep latest tag
+          for (const tag of tags.data) {
+            if (tag.name.startsWith('v')) {
+              if (tag.name.localeCompare(newest_tag.name, undefined, { numeric: true}) > 0) {
+                newest_tag = tag;
+              }
+            }
+          }
+          console.log(`Previous release has tag ${newest_tag.name} → ${newest_tag.commit.sha}`);
+
+          // Count all commits between HEAD and newest tag
+          // Limited to 250 commits
+          const distance = await github.rest.repos.compareCommitsWithBasehead({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            basehead: `${newest_tag.commit.sha}...${master_sha}`,
+          }).then(d => d.data.total_commits);
+
+          // Zip a zip file from dist directory
+          let release_name = `nightly_${distance}_${master_sha.substring(0,8)}` + '.zip';
+          console.log(`Release file name: ${release_name}`);
+          fs.renameSync('dist.zip', release_name);
+
+          // Upload the zip file to GitHub
+          const uploadedAsset = await github.rest.repos.uploadReleaseAsset({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            release_id: release.data.id,
+            name: release_name,
+            data: fs.readFileSync(release_name),
+            headers: {
+              'content-type': 'application/zip',
+            },
+          });
+
+          return uploadedAsset.data.browser_download_url;
+        result-encoding: string
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -335,3 +335,32 @@ jobs:
           target-folder: ${{ env.DOCS_VERSION_PATH }} # The folder the action should deploy to.
           commit-message: publish documentation
           clean: false # Releasing a new version is about creating a new directory, so we don't want to clean up the root.
+
+  delete-nightly-release:
+    runs-on: ubuntu-latest
+    needs: [deploy]
+
+    steps:
+      - name: Delete nightly release
+        uses: actions/github-script@v5
+        with:
+          github-token: ${{ secrets.gh_access_token }}
+          script: |
+            const releases = await github.rest.repos.listReleases({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+            })
+
+            const nightlyRelease = releases.data.find(r => r.tag_name === 'nightly'))
+
+            if (nightlyRelease) {
+              await github.rest.repos.deleteRelease({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                release_id: nightlyRelease.id,
+              })
+              console.log(`${nightlyRelease.tag_name} release has been deleted`)
+
+            } else {
+              console.log('No nightly release found')
+            }


dbarbier:
You made tests it in your repo but AFAICT not with this final version, please test and provide a link.
Code is complex, but if I am not mistaken, I emulated it on my repo between v0.9.2 and master (by moving randomly master) and it would look like https://github.com/dbarbier/scikit-decide/releases/tag/nightly (each 85MB file is replaced by dummy content to not waste resources), it is impossible to know which file is the latest one.

galleon:
> is
Indeed I did test it but get rid of most of my tests once `master` was reset.
The encoding of the name is made by git on `master` which has a linear history. In the following pattern `nightly-x-aha` the higher x the more recent is your file. 
It is therefore weird to have multiple similar x Can you detail how you did your tests?

dbarbier:
Consider that master and nightly are at the same commit. When master moves, `git describe --tags --long` will print something like `nightly-4-g<hash>`, then nightly tag is moved to master, and so forth.

dbarbier:
Even if master refers to the last release, it would not work better: https://github.com/dbarbier/scikit-decide/releases/tag/nightly2
Anyway I suggest to first publish only one asset, and when it works, to see how to keep multiple assets.

dbarbier:
What does `skdecide.__version__` look like?

galleon:
> @galleon What does `skdecide.__version__` look like?

with my current env:
➜  tmp git:(master) ✗ ipython                                                                                                                                                              
Python 3.8.6 (default, May 10 2021, 12:38:32)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.23.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import skdecide

In [2]: print(skdecide.__version__)
0.9.3-dev

Starting from 3.8 one could use `importlib.metadata.version` but we are still supporting 3.7

dbarbier:
Okay so this commit has nothing to do with nightly wheels and should be moved elsewhere.

galleon:
> Okay so this commit has nothing to do with nightly wheels and should be moved elsewhere.

You are right, it is connected as it can be used to track the version of an installed nightly wheel. It has been moved in an other PR.

galleon:
Versioning of nightly wheels has been changed to track them using:
   - distance to latest release (vx.y.z)
   - sha number

It has been tested on `galleon/master`:
   - using a `0, 1, *,  *, *` crontab
   - commiting dummy files

In `build.yml`:
  - if `nightly`tag does not exist it is created
  - Nighly release is created or updated
  - New artifact is added to the release

`release.yml` deletes the nightly tag and release

dbarbier:
Wheels are alphabetically sorted and not numerically sorted, which looks weird, see https://github.com/dbarbier/scikit-decide/releases/tag/nightly2

dbarbier (changes requested):
You assume everywhere that if nightly tag exists, then nightly release also exist (and vice-versa).  This is not robust, something may be broken during previous builds.  You must check against both tag and release when needed.

.github/workflows/build.yml
```diff
@@ -237,6 +241,131 @@ jobs:
...
+      - name: Create release
+        if: steps.nightly_tag.outputs.exists == '0'
+        uses: actions/github-script@v5
+        with:
+          github-token: ${{ secrets.gh_access_token }}
+          script: |
+            await github.rest.repos.createRelease({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              tag_name: 'nightly',
```
dbarbier, line 297: You could add: `prerelease: true`

.github/workflows/build.yml
```diff
@@ -2,8 +2,12 @@ name: 🔨 Build scikit-decide
 
 on:
   push:
+    tags-ignore:
+      - 'nightly'
```
nhuet, line 6: With the current filter in master `branches: - "**"` this is not needed as it will never be triggered on tags (which is the expected behaviour, only build for push on branches, not on any tags)

.github/workflows/build.yml
```diff
@@ -237,6 +241,131 @@ jobs:
...
+              release_id: release.data.id,
+              target_commitish: context.sha,
+            });
+
+      - name: Continuing only if master and nightly are different
+        if: github.sha != steps.nightly.outputs.sha
+        id: continue
+        run: |
+          echo "::set-output name=value::1"
+
```
nhuet, line 335: What is the purpose of this step? Maybe you wanted use it as a condition for each step below? But for now, they are just using the same condition ̀github.sha != steps.nightly.outputs.sha` so it looks like it is not used anywhere.

nhuet (commented):
I guess that notebooks/05_fake_tuto.ipynb is there for testing purpose and has to be removed before merging?

.github/workflows/build.yml
```diff
@@ -308,3 +308,190 @@ jobs:
           commit-message: publish documentation
           clean-exclude: |
             "version/*"
+
+  upload-nightly:
+    if: (github.ref == 'refs/heads/master') && (github.event_name == 'schedule')
```
nhuet, line 313: Did you forget to add the schedule event in "on" section ? For now it is triggering only on push and pull requests.

.github/workflows/build.yml
```diff
@@ -308,3 +308,190 @@ jobs:
...
+            } while (found === false && response.length > 0);
+            console.log(`Found ${distance} commits`);
+          } catch (err) {
+            console.log(err);
+          }
+
+          // Zip a zip file from dist directory
+          console.log(`nightly_${distance}_${ref_head.data.object.sha.substring(0,8)}.zip`);
+
+          let release_name = `nightly_${distance}_${ref_head.data.object.sha.substring(0,8)}` + '.zip';
```
nhuet, line 471: I would rather exchange the order and type `console.log(release_name)` to be sure to display what is actually used. (But not very important)

.github/workflows/build.yml
```diff
@@ -308,3 +308,190 @@ jobs:
...
+
+          fs.renameSync('dist.zip', release_name, function (err) {
+            if (err) throw err;
+            console.log('dist.zip was renamed to dist.zip');
```
nhuet, line 475: Not sure what this means. Do you meant
`console.log('dist.zip was renamed to ' + release_name);` ?

nhuet:
If I understand well this refactorying,
- each night, master having changed *or not*, we update the nightly tag to new master position and add the new wheels to assets while keeping all other ones as well
- at new release, we remove all these wheels to start again the process

Which means, between releases, we store wheels for all nights (e.g if there are 2 months between releases, we will store about 60 wheels zip). Am I correct ?

.github/workflows/build.yml
```diff
@@ -308,3 +310,195 @@ jobs:
...
+          try {
+            const tags = await github.rest.repos.listTags({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+            });
+
+            // Keep latest tag
+            for (const tag of tags.data) {
+              if (tag.name.startsWith('v')) {
+                if (tag.name > oldest_tag.name) {
```
dbarbier, line 434: This won't work (eg. v0.9.2 > v0.9.10), you need something as in versions.vue.

.github/workflows/build.yml
```diff
@@ -308,3 +310,195 @@ jobs:
...
+          let ref_head = null;
+          try {
+            ref_head = await github.rest.git.getRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'heads/master',
+            });
+          } catch (err) {
+            console.log(err);
+          }
```
dbarbier, line 345: try/catch not needed here

.github/workflows/build.yml
```diff
@@ -308,3 +310,195 @@ jobs:
...
+                if (commit.sha === oldest_tag.commit.sha) {
+                  found = true;
+                } else {
+                  if (found !== true) distance++;
+                }
+              });
+              page++;
+            } while (found === false && response.length > 0);
+            console.log(`Found ${distance} commits`);
+          } catch (err) {
```
dbarbier, line 471: if something went wrong we likely want to return or throw

.github/workflows/build.yml
```diff
@@ -308,3 +310,195 @@ jobs:
...
+              name: release_name,
+              data: fs.readFileSync(release_name),
+              headers: {
+                'content-type': 'application/zip',
+              },
+            });
+          } catch (err) {
+            console.log(err);
+          }
+          console.log(`Uploaded ${uploadedAsset.data.name} to GitHub`);
```
dbarbier, line 501: move this log into the try/catch clause, we do not want to display it if an error occurred

.github/workflows/release.yml
```diff
@@ -326,3 +326,38 @@ jobs:
...
+        uses: actions/github-script@v5
+        with:
+          github-token: ${{ secrets.gh_access_token }}
+          script: |
+            const releases = await github.rest.repos.listReleases({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+            })
+
+            const nightlyRelease = releases.data.find(r => r.tag_name.startsWith('nightly'))
```
dbarbier, line 345: why startsWith and not equality?

.github/workflows/release.yml
```diff
@@ -326,3 +326,38 @@ jobs:
...
+                release_id: nightlyRelease.id,
+              })
+              console.log(`${nightlyRelease.tag_name} release has been deleted`)
+
+              await github.rest.git.deleteRef({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                ref: `tags/${nightlyRelease.tag_name}`,
+              })
+              console.log(`${nightlyRelease.tag_name} tag has been deleted`)
```
dbarbier, line 360: why delete nightly tag?

dbarbier (changes requested):

dbarbier:
> If I understand well this refactorying,
> 
>     * each night, master having changed _or not_, we update the nightly tag to new master position and add the new wheels to assets while keeping all other ones as well
[...]
Nothing is done if master and nightly are at the same commit (test l. 369)

nhuet:
Actually @galleon added a commit for this test once I made the comment.

dbarbier:
> Actually the commit for this test was added once I made the comment.

okay this is another example why rebasing before review is finished is annoying.

nhuet:
This time at least, i do not think this was the issue as the commit has been done just after my comment and has not yet been squased or rebased (db0cb64fc56c61781791a7721ce593aadb6ccd48)

.github/workflows/build.yml
galleon: This has been changed using `localeCompare` with `numeric: true`

.github/workflows/build.yml
galleon: Changed

.github/workflows/release.yml
galleon: because there is no more `nightly` until the next PR

.github/workflows/build.yml
galleon: fixed

.github/workflows/build.yml
galleon: done

.github/workflows/release.yml
galleon: fixed

.github/workflows/build.yml
dbarbier: Good, I did not know about localeCompare, it works fine with multiple dots.  The sensitivity argument is useless here. Please reword the commit message, there is a typo localCompare -> localeCompare.

.github/workflows/release.yml
dbarbier: I still see it in your current branche (83b028a).  BTW commit d138345d does not do what it claims.

.github/workflows/release.yml
dbarbier: Deleting this tag or not won't change anything.  Less code less bugs.

.github/workflows/build.yml
dbarbier: agreed

.github/workflows/build.yml
```diff
@@ -308,3 +310,194 @@ jobs:
...
+        github-token: ${{ secrets.gh_access_token }}
+        script: |
+          const fs = require('fs');
+
+          // Get the ref for HEAD [A call to the API is probably useless]
+          let ref_head = await github.rest.git.getRef({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            ref: 'heads/master',
+          });
```
dbarbier, line 340: As mentioned in comment, there is no need for an API call, this could be written:
```js
    const master_sha = '${{ github.sha }}';
```
I renamed ref_head to master_sha because HEAD refers to the tip of any branch, not master.

.github/workflows/build.yml
```diff
@@ -308,3 +310,194 @@ jobs:
...
+          try {
+            ref_nightly = await github.rest.git.getRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'tags/nightly',
+            });
+
+            if (ref_nightly.data.object.sha === ref_head.data.object.sha) {
+              console.log('No new nightly release');
+              return;
```
dbarbier, line 354: No need for console.log, you can replace by
```js
    return "No new nightly release";
```

.github/workflows/build.yml
```diff
@@ -308,3 +310,194 @@ jobs:
...
+            tag_name: 'nightly',
+            name: 'nightly',
+            body: 'Nightly release crafted with ♥️ somewhere on 🌎',
+            draft: false,
+            prerelease: true,
+          });
+          console.log(`Updated ${release.data.tag_name} nightly release  ${release.data.draft} ${release.data.prerelease}`);
+
+          // Get all tags and keep the oldest one starting by v
+          let oldest_tag = { name: 'v0.0.0' };
```
dbarbier, line 418: Replace oldest by newest everywhere.

.github/workflows/build.yml
```diff
@@ -308,3 +310,194 @@ jobs:
...
+          console.log(`Release does exist with tag ${release.data.tag_name} [${release.data.draft} ${release.data.prerelease}]`);
+
+          // At this stage both tag & release exist
+
+          // Update nightly tag
+          await github.rest.git.updateRef({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            ref: 'tags/nightly',
+            sha: ref_head.data.object.sha,
```
dbarbier, line 400: It would be wise to add `force: true` in case master has been rebased non fast-forward.

.github/workflows/build.yml
dbarbier: I found a simpler solution:
```js
       // Limited to 250 commits
       const distance = await github.rest.repos.compareCommitsWithBasehead({
         owner: context.repo.owner,
         repo: context.repo.repo,
         basehead: `${oldest_tag.commit.sha}...${master_sha}`,
       }).then(d => d.data.total_commits);
```
If there are more than 250 commits, the list of nighly wheels is not manageable and I am not sure that having wheels with the same distance won't make any significant difference.

.github/workflows/build.yml
```diff
@@ -308,3 +310,194 @@ jobs:
...
+
+          // Zip a zip file from dist directory
+          console.log(`nightly_${distance}_${ref_head.data.object.sha.substring(0,8)}.zip`);
+
+          let release_name = `nightly_${distance}_${ref_head.data.object.sha.substring(0,8)}` + '.zip';
+
+          fs.renameSync('dist.zip', release_name, function (err) {
+            if (err) throw err;
+            console.log(`dist.zip was renamed to {release_name}`);
+          });
```
dbarbier, line 481: Why not simply
```
    fs.renameSync('dist.zip', release_name);
```
?  Less code less bugs.

.github/workflows/build.yml
```diff
@@ -308,3 +310,194 @@ jobs:
...
+                // if (tag.name > oldest_tag.name) {
+                if (tag.name.localeCompare(oldest_tag.name, undefined, { numeric: true, sensitivity: 'base'}) > 0) {
+                  oldest_tag = tag;
+                }
+              }
+            }
+          } catch (err) {
+            console.log(err);
+            throw new Error('Failed to get tags');
+          }
```
dbarbier, line 438: There are several occurrences of
```js
try {
  ...
} catch(err) {
  console.log(err)
  throw new Error(...);
}
```
This is useless, remove the try/catch, the error message will already get displayed in log.

.github/workflows/release.yml
galleon: This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.

galleon: Same for the second one

dbarbier: They are on your fork, and displayed when you click on the links; I wall update them anyway

.github/workflows/build.yml
galleon: done

.github/workflows/build.yml
galleon: Added

.github/workflows/build.yml
galleon: Yes ! done

.github/workflows/build.yml
galleon: Indeed. I learned something

.github/workflows/release.yml
galleon: Removed

.github/workflows/build.yml
galleon: done

.github/workflows/build.yml
galleon: done

.github/workflows/build.yml
galleon: done

.github/workflows/build.yml
galleon: done

.github/workflows/build.yml
galleon: done

nhuet (approved):
LGTM. Think to squash the commits before merging.

.github/workflows/release.yml
dbarbier: still not fixed

.github/workflows/build.yml
```diff
@@ -343,3 +345,156 @@ jobs:
...
+          let release = null;
+          try {
+            release = await github.rest.repos.getReleaseByTag({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              tag: 'nightly',
+            });
+            console.log(`Found release ${release.data.tag_name} ${release.data.draft} ${release.data.prerelease}`);
+          } catch (err) {
+            console.log(`Release ${tag} not found`);
```
dbarbier, line 406: variable 'tag' not defined

.github/workflows/build.yml
```diff
@@ -343,3 +345,156 @@ jobs:
...
+
+          const tags = await github.rest.repos.listTags({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+          });
+
+          // Keep latest tag
+          for (const tag of tags.data) {
+            if (tag.name.startsWith('v')) {
+              // if (tag.name > newest_tag.name) {
```
dbarbier, line 459: remove dead code

.github/workflows/build.yml
```diff
@@ -343,3 +345,156 @@ jobs:
...
+              if (tag.name.localeCompare(newest_tag.name, undefined, { numeric: true, sensitivity: 'base'}) > 0) {
```
dbarbier, line 460: 'sensitivity' argument is useless

.github/workflows/build.yml
```diff
@@ -343,3 +345,156 @@ jobs:
...
+          }).then(d => d.data.total_commits);
+
+          // Zip a zip file from dist directory
+          let release_name = `nightly_${distance}_${master_sha.substring(0,8)}` + '.zip';
+          console.log(`Release file name: ${release_name}`);
+          fs.renameSync('dist.zip', release_name);
+
+          // Upload the zip file to GitHub
+          let uploadedAsset = null;
+          uploadedAsset = await github.rest.repos.uploadReleaseAsset({
```
dbarbier, line 487: both lines can be merged: const uploadAsset =...

.github/workflows/build.yml
dbarbier: in fact this log is useless, this string is already in job output

dbarbier (approved):

PR #98, opened 2021-11-08
Keep track of version in library

galleon:
This PR does the following:
- request poetry to change version in `skdecide/__iniy__.py`
- print the information when `utils/collect_env.py`is called

### DIFF
--- a/utils/collect_env.py
+++ b/utils/collect_env.py
@@ -18,6 +18,11 @@ try:
 except (ImportError, NameError, AttributeError, OSError):
     TORCH_AVAILABLE = False
 
+try:
+    from skdecide import __version__ as skdecide_version
+except (ImportError):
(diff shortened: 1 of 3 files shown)


dbarbier:
Please write in commit message the new version number, it is hard to guess.

utils/collect_env.py
```diff
@@ -18,6 +18,12 @@
 except (ImportError, NameError, AttributeError, OSError):
     TORCH_AVAILABLE = False
 
+try:
+    from skdecide import __version__
```
dbarbier, line 22: You could write
```python
try:
    from skdecide import __version__ as skdecide_version
 except (ImportError):
    skdecide_version = "not available"
```
and use `skdecide_version ` directly

dbarbier (approved):

PR #100, opened 2021-11-09
[trivial] Apply pre-commit linters on skdecide/

dbarbier:
The only open pull request which modifies files under skdecide/
directory is #98, conflict resolution will be trivial, so we can apply
pre-commit linters.

There is no code change, this commit only contains changes performed
by running

    pre-commit run -a


dbarbier:
Branch rebased to resolve conflicts.

PR #114, opened 2021-11-21
Re-add dropdown menu for versioned docs

dbarbier:
It had been erroneously removed by 7e30b12.
There is no reason to set its class to "can-hide".


dbarbier:
Superseded by #115.

PR #115, opened 2021-11-21
Fixing multiple problems in doc

galleon:
- typo in workflow
- re-instating version tag in navbar
- typo in name of default branch

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -302,7 +302,6 @@ jobs:
         if: github.ref == 'refs/heads/master'
         uses: JamesIves/github-pages-deploy-action@4.1.5
         with:
-          token: ${{ secrets.gh_access_token }}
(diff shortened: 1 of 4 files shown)


dbarbier (commented):
Commit message of b508bca is misleading, it seems to imply that there is a bug to fix, but this is not the case. This code works with any label; you can change "main" to "master" if you want, but this is cosmetics.

dbarbier (approved):

dbarbier:
In fact it would be better to drop the "token" option of deploy, it is useless and may break if there is an error in secrets.

dbarbier:
You may also drop "tokem" from release.yml ;-)

galleon:
> You may also drop "tokem" from release.yml ;-)

Oops :-) Fixed

dbarbier:
Looks good.

PR #116, opened 2021-11-22
Initiate a contribute page to online documentation

nhuet:
- a link is available in top-left menu
- it give info about
  - installation from source in dev mode
  - local build of the docs
  - notebooks convention to extract title and description
  - unit tests

For now, it still misses guidelines about how to actually contribute.
So next steps (for further PRs) could be
- include info from CONTRIBUTING.md in root directory, that then could point to online doc
- update code style convention by adding how to use pre-commit to lint the code

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -45,6 +45,10 @@ module.exports = {
                 text: 'Reference',
                 link: '/reference/'
             },
+            {
+                text: 'Contribute',
+                link: '/contribute'
+            },
         ],
         sidebar: 'auto',
         markdown: {
--- a/docs/contribute.md
+++ b/docs/contribute.md
@@ -1,0 +1,193 @@
+# Contributing to scikit-decide
+
+This page is intended to help people wanting to contribute to the library.
+
+For now is it mainly about how to test locally changes made to the library.
+In the future, it will also cover guidelines to follow when contributing.
+
+## Installing from source in developer mode
+
+> **Disclaimer**: The following process has only been tested on Linux/MacOS platforms.
+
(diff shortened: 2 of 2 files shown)


galleon:
I think it is a good idea to have named it contribute. It might therefore be the opportunity to describe briefly the various tools that are run of a PR and describe what are the best practices to run some of those tools locally before committing. What do you think ?

galleon:
I think it is a good idea to have named it contribute. It might therefore be the opportunity to describe briefly the various tools that are run of a PR and describe what are the best practices to run some of those tools locally before committing. What do you think ?

nhuet:
Exactly, that's the idea. Even though it is not yet the case, we should add later stuff about good practice for PRs.

dbarbier (approved):

galleon:
Can you please add a preview ? I think adding a part on how to check code with linters locally is a must since we have it now. @dbarbier can you please add something ?

dbarbier:
> I think adding a part on how to check code with linters locally is a must since we have it now. @dbarbier can you please add something ?

It will be simpler for me to make a PR after this one is merged.

nhuet:
> Can you please add a preview ? 

You mean a screenshot of resulting doc ? Here you are:

Screenshot from 2021-11-25 10-16-30


For full doc, download the artifact named "doc" from the workflow triggered by this PR, then
```shell
mkdir testdoc
cd testdoc
unzip path/to/doc.zip -d scikit-decide
python -m http.server
```

nhuet:
I proposed a more complete contributing guide in #123 . This includes a section about code linters. If the PR #123 suits you, we can drop this one.

PR #120, opened 2021-11-23
Add a quickstart notebook made of previous small autogenerated notebooks

nhuet:
As discussed, I added a first notebook that gather the previous ones that were generated from some of the python scripts in examples/ folder. I omitted the one about the maze which seems to be just a lighter version of the maze tutorial already existing.

Now that I have done it, i am not really convinced because:

- I find it not so easy to understand from a newbie point of view.
- That makes duplicates with existing notebooks. All sections are more or less already sections of the gym tutorial.
- As it is, the notebook is not working on binder because of the way gym displays its environments. We can
  -  either create our own rollout as in gym tutorial => but this is will complexify the notebook and miss the purpose of quickness
  - or embed the environment in a monitor wrapper to generate movies

So to make it understandable and working on binder, some extra stuff should be added, and it seems to me that doing so it will looks very much like the existing gym tutorial.

### DIFF
--- a/notebooks/01_quickstart.ipynb
+++ b/notebooks/01_quickstart.ipynb
@@ -1,0 +1,273 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "484323a5",
+   "metadata": {},
+   "source": [
+    "# Getting started\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


neo-alex:
OK, I see your points and I am fine with not including this notebook in the documentation (as long as we keep these quick examples in the [examples](https://github.com/airbus/scikit-decide/tree/master/examples) folder of the repo).

PR #125, opened 2021-11-26
Add benchmarking tutorial

nhuet:
This notebook shows how to use ray tune library to benchmark several solvers.

This PR is taking over PR #93 and adapt the benchmarking notebook. The main modifications are:
- pass a specific progress_reporter for jupyter to `tune.run` so that progress is updated in place
- display an html version of the dataframe instead of using bare `print`
- remove the tensorboard part which is not working on binder and finally not so relevant as there are only 1 point by solver and only 2 columns in the data really interesting
- add some explanations when needed

[Binder](https://mybinder.org/v2/gh/airbus/scikit-decide/binder?urlpath=git-pull%3Frepo%3Dhttps%253A%252F%252Fgithub.com%252Fnhuet%252Fscikit-decide%26urlpath%3Dtree%252Fscikit-decide%252Fnotebooks%252F14_benchmarking_tuto.ipynb%26branch%3Dnh%252Fbenchmarking_nb_2)

### DIFF
--- a/notebooks/14_benchmarking_tuto.ipynb
+++ b/notebooks/14_benchmarking_tuto.ipynb
@@ -1,0 +1,297 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Benchmarking scikit-decide solvers\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


neo-alex (approved):
All good for me, LGTM

PR #126, opened 2021-11-29
Update notebook's kernel name and display name for default one (python3)

nhuet:
The current chosen kernel name (cinb_env) was specific to my configuration and it will be not found on a generic computer.

It results to
- a popup message asking for choosing the appropriate kernel when running locally (so not completely an error but still unsatisfactory)
- an error when testing the notebook with tool like nbmake with `pytest --nbmake`

We may want to add a pre-commit hook to check that later.

### DIFF
--- a/notebooks/14_benchmarking_tuto.ipynb
+++ b/notebooks/14_benchmarking_tuto.ipynb
@@ -257,9 +257,9 @@
  ],
  "metadata": {
   "kernelspec": {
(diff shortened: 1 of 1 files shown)


dbarbier:
Why is this change needed? Please explain in the commit message what issue it solves.

nhuet:
Commit message and PR updated with the appropriate info.

dbarbier (approved):

PR #127, opened 2021-11-29
Trigger build  workflow for all branches

nhuet:
With previous syntax, the workflow did not trigger on push on
branches with a slash in its name (e.g. nh/ci_all_branches).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,7 @@ name: 🔨 Build scikit-decide
 on:
   push:
     branches:
-      - "*"
+      - "**"
   pull_request:
   workflow_dispatch:
 


dbarbier (approved):

galleon:
But this will also trigger build on tags on master which is not desired. One need probably to add `- !'v[0-9]+.[0-9]+.[0-9]+'`

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

dbarbier:
> But this will also trigger build on tags on master [...]

No

nhuet:
> But this will also trigger build on tags on master which is not desired. One need probably to add `- !'v[0-9]+.[0-9]+.[0-9]+'`
> 
> https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

This will not. We had already `branches = "*"` which triggered on all branches that did not have a "/" in the name, but not on tags. 
With `branches = "**"`, it triggers on all branches even if ther is a "/" but still not on tags as the "tags" pattern as not been filled.
Cf  https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestbranchestags: 
> If you define only tags or only branches, the workflow won't run for events affecting the undefined Git ref.

Issue #121, opened 2021-11-25
Manage notebooks in CI


galleon:
## 🚀 Feature
Manage notebooks in CI

## Motivation
Since the documentation rely on notebooks, they need to be tested as any other artifact

## Pitch
We need to make sure that each notebook run without any problem with the current code. Notebooks can be considered as tests and might sometimes required specific packages to run (on top of the scikit-decide library) i.e. additional test packages will have to be added to `poetry` or a `notebook-requirements.txt`

## Alternatives

None

## Additional context

- [nbmake](https://github.com/treebeardtech/nbmake)
- [post](https://semaphoreci.com/blog/test-jupyter-notebooks-with-pytest-and-nbmake)

nhuet:
- **Locally**: To do it locally, one can install nbmake and then `pytest --nbmake notebooks/*.ipynb`.
In [nh/test_notebooks](https://github.com/nhuet/scikit-decide/tree/nh/test_notebooks), nbmake is added as dev dependency in poetry so that `poetry run pytest --nbmake notebooks/*.ipynb` works.
- **in CI**: 
   - We can do the same (install nbmake + pytest command) except we need extra dependencies to make gym nb work (ipyvirtual display, pyopengl, xvfb) --> this is done in [nh/ci_test_nb](https://github.com/nhuet/scikit-decide/tree/nh/ci_test_nb) by adding requirements files in notebooks/ folder, only for unix test to avoid complexifying dependencies to manage according to env (and on binder, it will only use unix worker). *[wip]*
   - Eventually, we could use the binder env generated (perhaps by #122) to test notebooks as the purpose is mainly to ensure that the notebooks will run on lbinder.

Note: we have to check that testing times remain reasonable, perhaps by specifying a timeout by cell. Nbmake allows that but generates an error when timeout is reached. We can also tell it to not raise errors but then we miss the purpose of testing notebooks... So light versions of notebooks are needed, perhaps by parametriziing them using [papermill ](https://papermill.readthedocs.io/en/latest/)

g-poveda:
Decision : we want to focus more on the implementation of more unit test.

PR #128, opened 2021-11-29
Add option to strip kernelspec metadata from notebooks with pre-commit

nhuet:
**Goal**: Avoid having weird non default kernels specified by accident in notebooks.

**Context**: If an unknown kernel is specified, local jupyter sessions will popup to ask for selecting among existing kernels. Testing notebooks with `pytest --nbmake` will crash.

**Pros**:
When removing the entire field "kernelspec" from metadatas, the notebooks will open in jupyter with the default kernel which is python3. This will also work with `pytest --nbmake` when testing notebooks.

**Cons**:
When saving the notebooks from a local Jupyter session, the kernelspec metadata will be restored. So the pre-commit will probably complain after each change if it has been installed as a git hook. So that pre-commit will need to be performed after each change.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -60,4 +60,5 @@ repos:
     rev: 0.5.0
     hooks:
       - id: nbstripout
+        args: [--extra-keys=metadata.kernelspec]
         files: ^notebooks/
--- a/notebooks/11_maze_tuto.ipynb
+++ b/notebooks/11_maze_tuto.ipynb
@@ -685,11 +685,6 @@
  ],
  "metadata": {
   "anaconda-cloud": {},
-  "kernelspec": {
-   "display_name": "Python 3",
-   "language": "python",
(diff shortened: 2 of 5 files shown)


nhuet:
Cf https://github.com/kynan/nbstripout/issues/141

dbarbier:
IMO cons outweigh pros.

nhuet:
Ok. Another solution could be to modify inplace the name of the kernel for python3. And yet another one to trust people will not do as i did.

Actually, once the notebooks will be tested during CI, we will have a test failing thus it is ok for me. (cf #121)

dbarbier:
> Ok. Another solution could be to modify inplace the name of the kernel for python3.[...]

That would be safer, yes.

alex-treebeard:
@nhuet thanks for raising the crash bug. Will fix in the next week https://github.com/treebeardtech/nbmake/issues/37

PR #131, opened 2021-11-30
Create specific binder env for each scikit-decide release

nhuet:
**Summary**:
During release.yml workflow, we push a new tag  defining the corresponding binder environment, then build and cache it on mybinder.org. Finally the binder links in the release doc point to this specific environment.

**Hypotheses**:
- a branch named "binder" exists on the repository with an environment.yml file  used to define a binder env
- the branch is up-to-date with necessary dependencies to make work the notebooks on master with master version of scikit-decide
- environment.yml file contains the dependency in scikit-decide, and it follows the format `scikit-decide[all]==...`

**Process**:
- we checkout binder branch
- we replace the spec for scikit-decide by the scikit-decide version corresponding to the release triggering the workflow
- we commit and push a tag named "binder_vx.x.x" where x.x.x stands for the inferred version number
- we force a build of the environment on mybinder.org to accelerate later call to it
- we use this binder env for links in the doc for the release

Note:  we used a quite verbose way to define the BINDER_TAG variable. This is to make it clear that we create it exactly the same way in both jobs "create-binder-env" and "build-docs" , as there is no simple way to share variables between jobs (without using artifacts).

A fake release v0.9.3 has been created on nhuet/scikit-decide to test that the binder environment is well defined. 
When the workflow will be finished, we should have a tag "binder_v0.9.3" (see https://github.com/nhuet/scikit-decide/tree/binder_v0.9.3 ) defining the binder environment with scikit-decide 0.9.3 (from pypi), and the environment should have been prebuilt on mybinder.org. Test this link: [Binder](https://mybinder.org/v2/gh/nhuet/scikit-decide/binder_v0.9.3?urlpath=git-pull%3Frepo%3Dhttps%253A%252F%252Fgithub.com%252Fnhuet%252Fscikit-decide%26urlpath%3Dtree%252Fscikit-decide%252Fnotebooks%252F11_maze_tuto.ipynb%26branch%3Dv0.9.3)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -303,8 +303,7 @@ jobs:
 
       - name: Update scikit-decide dependency and tag the environement
         run: |
-          sed -e "s/scikit-decide\[all\]==.*/scikit-decide\[all\]==${SKDECIDE_VERSION}/" environment.yml > environment.yml.new
-          mv environment.yml.new environment.yml
+          sed -i -e "s/scikit-decide\[all\]==.*/scikit-decide[all]==${SKDECIDE_VERSION}/" environment.yml
           git config user.name "Binder Bot"
           git config user.email "binder@bot.com"
           git commit environment.yml -m "Update scikit-decide dependency"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -283,63 +283,52 @@ jobs:
           packages_dir: wheels/
           repository_url: https://test.pypi.org/legacy/
 
-  create-binder-env:
-    needs: [deploy]
-    runs-on: ubuntu-latest
-    steps:
-      - name: Get scikit-decide release version and define new binder env tag
-        run: |
-          TAG_NAME=${GITHUB_REF/refs\/tags\//}  # stripping "refs/tags/"
-          SKDECIDE_VERSION=${TAG_NAME/v/}  # stripping "v"
-          BINDER_TAG=binder_${TAG_NAME}
-          echo "BINDER_TAG=${BINDER_TAG}" >> $GITHUB_ENV
-          echo "SKDECIDE_VERSION=${SKDECIDE_VERSION}" >> $GITHUB_ENV
-
-      # Checks-out binder branch
-      - name: checkout binder branch
-        uses: actions/checkout@v2
-        with:
-          ref: binder
-
-      - name: Update scikit-decide dependency and tag the environement
-        run: |
-          sed -i -e "s/scikit-decide\[all\]==.*/scikit-decide[all]==${SKDECIDE_VERSION}/" environment.yml
-          git config user.name "Binder Bot"
-          git config user.email "binder@bot.com"
-          git commit environment.yml -m "Update scikit-decide dependency"
-          git tag -a ${BINDER_TAG} -m "Release binder environment for scikit-decide ${SKDECIDE_VERSION}"
-          git push origin ${BINDER_TAG}
-
-      - name: Cache binder build on mybinder.org
-        uses: jupyterhub/repo2docker-action@master
-        with:
-          NO_PUSH: true
-          MYBINDERORG_TAG: ${{ env.BINDER_TAG }}
-
   build-docs:
-    needs: [create-binder-env]
+    needs: [deploy]
     runs-on: ubuntu-latest
     env:
       DOCS_VERSION_PATH: /
 
     steps:
-      - name: Set env variables for github+binder links in doc
+      - name: Get scikit-decide release version and update online docs path
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -278,19 +278,55 @@ jobs:
...
+      # Checks-out binder branch
+      - name: checkout binder branch
+        uses: actions/checkout@v2
+        with:
+          ref: binder
+
+      - name: Update scikit-decide dependency and tag the environement
+        run: |
+          sed -e "s/scikit-decide\[all\]==.*/scikit-decide\[all\]==${SKDECIDE_VERSION}/" environment.yml > environment.yml.new
+          mv environment.yml.new environment.yml
```
dbarbier, line 302: Replace these 2 lines by
```
sed -i -e "s/scikit-decide\[all\]==.*/scikit-decide[all]==${SKDECIDE_VERSION}/" environment.yml
```

.github/workflows/release.yml
```diff
@@ -278,19 +278,55 @@ jobs:
...
+          git config user.name "Binder Bot"
+          git config user.email "binder@bot.com"
```
dbarbier, line 304: Which value is set if user.email empty?  A github one, or the mail from the person who pushed the tag?

.github/workflows/release.yml
```diff
@@ -278,19 +278,55 @@ jobs:
...
+          git commit environment.yml -m "Update scikit-decide dependency"
+          git tag -a ${BINDER_TAG} -m "Release binder environment for scikit-decide ${SKDECIDE_VERSION}"
+          git push origin ${BINDER_TAG}
+
+      - name: Cache binder build on mybinder.org
```
dbarbier, line 309: Why do we want to create this cache?

.github/workflows/release.yml
```diff
@@ -278,19 +278,55 @@ jobs:
...
+  build-docs:
+    needs: [create-binder-env]
+    runs-on: ubuntu-latest
     env:
       DOCS_VERSION_PATH: /
 
     steps:
       - name: Set env variables for github+binder links in doc
         run: |
+          TAG_NAME=${GITHUB_REF/refs\/tags\//}  # stripping "refs/tags/"
```
dbarbier, line 324: TAG_NAME could be put into GITHUB_ENV

.github/workflows/release.yml
nhuet: To avoid a long waiting time for the first user who will click on the links in the doc. It is not a big deal that the release workflow takes a bit more time imho (as it is for release and not each push)

.github/workflows/release.yml
nhuet: What for ? Maybe for DOCS_VERSION_PATH ? (i did not see that i could reuse something here also) But it should be SKDECIDE_VERSION in this case

.github/workflows/release.yml
nhuet: From my test it creates it automatically and it gives something like "runner@fv-az36-926.1s0p3okewbbu3agytf0rxb3dlg.gx.internal.cloudapp.net"

.github/workflows/release.yml
nhuet: I will take care of binder cache in an other PR. The way it is done in repo2docker-action is perfectible and it will complexify this PR for nothing

nhuet:
I took into account @dbarbier comments. I force rebased in order to take account the correction that make doc building works (fe50cb9dbd324327a5a8f21a3e7b0a3ab2273427).

dbarbier (approved):
Looks good. Did you test it on your repo after these changes?

nhuet:
Yes, i retag a v0.9.3 on my fork. The binder env has been generated (but is now on a binder_test branch as i rebased my binder branch for PR #139) and my online doc is pointing on the proper sha1. See notebooks page here: https://nhuet.github.io/scikit-decide/version/0.9.3/notebooks/

nhuet:
Actually i just tried to follow the links in doc (thus pointing to the binder env corresponding to the appropriate sha1), and
- it works
- but it rebuild each time => does that mean that pointing to a sha1 prevent binder to understand that it has already been built?
- the second build is still very short because docker has a lot of things in cache => this is because i did it in a row, but i am not sure tomorrow it will be the same. While pointing to a branch or a tag (instead of a sha1) seemed to make binder recognize it as already built and only pull the appropriate docker image from the hub.

nhuet:
> * but it rebuild each time => does that mean that pointing to a sha1 prevent binder to understand that it has already been built?

Nevermind. I tried a 3rd times, and it found the build and only pulled the image. Maybe i was unlucky. It may be due to the fact that mybinder.org is actually deploying binderhub on several cloud infrastructure (cf https://mybinder.readthedocs.io/en/latest/about/federation.html#the-binderhub-federation)

PR #132, opened 2021-12-02
Choosing node version explicitely

galleon:
Node version is forced to 14 to avoid doc generation crash

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -280,6 +280,10 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -219,6 +219,10 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+
```
nhuet, line 225: should be in build-docs, not in test-windows

nhuet (changes requested):

.github/workflows/build.yml
galleon: Yep ... fixed !

dbarbier (approved):

PR #136, opened 2021-12-06
Fix libminizinc build when ccache is installed

dbarbier:
CMake 3.4 added native support for ccache and distcc, by setting CMAKE_\<lang>_COMPILER_LAUNCHER variables.

But libminizinc option USE_CCACHE is harmful, it adds ccache again to the command line, and there is a build error:
```
ccache: error: Recursive invocation (the name of the ccache binary must be "ccache")
```
We did not see it in previous build because the cache on build dependencies was hit.


dbarbier:
@fteicht To answer your question about build speedup, the build steps on ubuntu and macos took 2-3 minutes instead of 12-15 minutes here.  Timings of other steps are unchanged.

nhuet (approved):

PR #137, opened 2021-12-06
Test notebooks locally

nhuet:
- Add nbmake in dev dependencies for testing notebooks locally with pytest
- Add explanation in contributing guide.

### DIFF
--- a/docs/contribute.md
+++ b/docs/contribute.md
@@ -188,6 +188,14 @@ From the "scikit-decide" root directory, run unit tests with:
  poetry run pytest tests
 ```
 
+### Running notebooks as tests
+
+One can test programmatically that notebooks are not broken thanks to [nbmake](https://github.com/treebeardtech/nbmake) extension for pytest.
+
+```shell
+poetry run pytest --nbmake notebooks -v
+```
+
 ## Guidelines to follow when preparing a contribution
 
 ### Coding style and code linting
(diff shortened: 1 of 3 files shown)


nhuet:
To test the PR, you have to install again with poetry (i do not know how to update the poetry install cleanly). And then type
`poetry run pytest --nbmake notebooks -v`

nhuet:
I see that the workflow crashed but is again a seg fault for tests/solvers/cpp that happens from time to time with ubuntu or macos runners (here ubuntu). It would be nice to find out why this is happening. And if someone with the proper rights (@galleon ?) can relaunch the workflow, it should work normally... (it worked on my branch that this PR is pulling: https://github.com/nhuet/scikit-decide/actions/runs/1544104432)

nhuet:
Thanks @galleon ! So now that the workflow passed, what is missing for the merge ?

galleon:
Still need to move `nbmake` as a dev dependency.

nhuet:
Weird I was sure to have added it to dev section... I move it in pyproject.toml then reinstall everything (:cry:) so that poetry.lock update accordingly and i repush...

nhuet:
I am not sure if I am doing it properly. To add a dependency while having a poetry.lock, I
- added nbmake to dev dependencies in pyproject.toml
- run a `poetry update` so that the new dependencies are taken into account (else the poetry install fails)

=> it results in all versions of dependencies being recomputed and updated to newest available (not necessarily what we wanted). Maybe I could have first done a `poetry update` in a separate commit, before adding the new dep nbmake...

Anyway, it should work like that, let us see what the workflow give us after this update...

PR #139, opened 2021-12-07
Binder env: install scikit-decide from last nightly build 

nhuet:
We remove scikit-decide dependency from environment.yml so that the installation is done in postBuild:
- only if scikit-decide not yet installed (so that release env are not  affected as they will have scikit-decide specified in environment.yml)
- we use github rest api and jq parser to extract download url for last nightly build from "nightly" release
- if not available (no tag or no asset), we fall back to plain `pip install scikit-decide` which will pull from pypi
- else we unzip and install appropriate wheel

Notes: 
- for now we download the nightly build from galleon/scikit-decide, but we will switch to airbus/scikit-decide once the first nightly release is done there.
- one can test the resulting environemnt here: https://mybinder.org/v2/gh/nhuet/scikit-decide/binder then open a notebook and test `import skdecide; skdecide.__version__` in a cell.
- in order to have really the *last* nightly build, we have to force a rebuild of this branch ("binder") when building a new nightly release. This will be done in a new PR by modifying master's build.yml.  The only way is by making a push to the "binder" branch. (see [below](https://github.com/airbus/scikit-decide/pull/139#issuecomment-987942953))

### DIFF
--- a/postBuild
+++ b/postBuild
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Last nightly build
+# Last nightly build: No new nightly release
 NIGHTLY_TAG="nightly"
 NIGHTLY_REPO="galleon/scikit-decide"
 
--- a/postBuild
+++ b/postBuild
@@ -1,23 +1,26 @@
 #!/bin/bash
 
+# Stops if an error occurs
+set -e
+
 # Last nightly build: No new nightly release
 NIGHTLY_TAG="nightly"
 NIGHTLY_REPO="galleon/scikit-decide"
 
 # Install scikit-decide from last nightly build if not yet installed
-skdecide_version=$(pip list | grep "scikit-decide")
+skdecide_version=$(pip list | grep "scikit-decide" | head -n 1)
 if [ -z "${skdecide_version}" ]; then
   echo "Scikit-decide not found: installing from last nightly build."
   # get last nightly build asset url
   asset_url=$(\
     curl -s https://api.github.com/repos/${NIGHTLY_REPO}/releases \
     | jq -c ".[] | select( .tag_name == \"$NIGHTLY_TAG\" ) | .assets|sort_by(.createdAt)|.[-1].browser_download_url" \
-    | sed -e 's/\"//g'
+    | sed -e 's/"//g'
   )
   # check that it exists (either "null" is no asset in nightly or "" if no tag nightly at all)
   if  [ $asset_url == "null" ] || [ $asset_url == "" ]; then
     echo "No nightly build available. Falling back to last version in pypi."
-    pip install sickit-decide[all]
+    pip install scikit-decide[all]
   else
     # download and unzip
     wget --output-document=nightly.zip ${asset_url}
--- a/postBuild
+++ b/postBuild
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# Last nightly build
+NIGHTLY_TAG="nightly"
+NIGHTLY_REPO="galleon/scikit-decide"
+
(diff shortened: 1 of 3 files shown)


nhuet:
Actually, this not doing what we really want: after the first build (first access to the environemnt), the docker image is pushed to dockerhub so that it will not be rebuilt even if a nightly build occurs. We still need to trigger a new build when updating the nightly release.

postBuild
```diff
@@ -0,0 +1,33 @@
...
+
+# Install scikit-decide from last nightly build if not yet installed
+skdecide_version=$(pip list | grep "scikit-decide")
+if [ -z "${skdecide_version}" ]; then
+  echo "Scikit-decide not found: installing from last nightly build."
+  # get last nightly build asset url
+  asset_url=$(\
+    curl -s https://api.github.com/repos/${NIGHTLY_REPO}/releases \
+    | jq -c ".[] | select( .tag_name == \"$NIGHTLY_TAG\" ) | .assets|sort_by(.createdAt)|.[-1].browser_download_url" \
+    | sed -e 's/\"//g'
```
dbarbier, line 15: backslash not needed

postBuild
```diff
@@ -0,0 +1,33 @@
...
+  )
+  # check that it exists (either "null" is no asset in nightly or "" if no tag nightly at all)
+  if  [ $asset_url == "null" ] || [ $asset_url == "" ]; then
+    echo "No nightly build available. Falling back to last version in pypi."
+    pip install sickit-decide[all]
```
dbarbier, line 20: typo

postBuild
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
```
dbarbier, line 2: I do not know if there are guidelines for writing postBuild scripts, but is is good practice in general for such scripts to add
```shell
set -e
```
at the beginning of the script so that it aborts if any command fails.

postBuild
```diff
@@ -0,0 +1,33 @@
...
+# Last nightly build: No new nightly release
+NIGHTLY_TAG="nightly"
+NIGHTLY_REPO="galleon/scikit-decide"
+
+# Install scikit-decide from last nightly build if not yet installed
+skdecide_version=$(pip list | grep "scikit-decide")
```
dbarbier, line 8: test below will crash if for any reason skdecide_version contains a newline character; add `| head -n 1` to make sure that it contains a single line

postBuild
```diff
@@ -0,0 +1,33 @@
...
+    pip install sickit-decide[all]
+  else
+    # download and unzip
+    wget --output-document=nightly.zip ${asset_url}
+    unzip nightly.zip
+    # get the appropriate wheel from python version
+    wheel_pythonversion_tag=$(python -c 'import sys; print(f"cp{sys.version_info.major}{sys.version_info.minor}")')
+    wheel_path=$(ls dist/scikit_decide*${wheel_pythonversion_tag}*manylinux*.whl)
+    # install scikit-decide with all extras
+    pip install ${wheel_path}[all]
```
dbarbier, line 29: in interactive shells, "[all]" must be quoted so that brackets are not interpreted by bash, not sure if this is required here too.

dbarbier:
Never mind, I think now that we should postpone this idea until we have a real use case to test these changes.

postBuild
nhuet: it worked like that in my tests

nhuet:
As we want to wait for this PR and as i need a working "binder" branch on my repo, i close this PR. I should have use another branch name :) . The work is saved now under nh/binder_nightly branch. I will reopen a PR later pointing to this new branch.

PR #140, opened 2021-12-07
Force a rebuild of binder environment at each nightly build

nhuet:
This is done by pushing a "dummy" commit to binder branch.

The commit only updates a comment with the nightly build name. The comment itself does nothing, but the fact that the binder branch is updated will trigger a new build the next time the binder environment will be called.

In a later PR, we could also trigger manually the build on mybinder.org in order to cache the build and reduce starting time for next users. But I prefer to dissociate the two features.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -381,7 +381,7 @@ jobs:
             });
 
             if (ref_nightly.data.object.sha === master_sha) {
-              return "No new nightly release";
+              return '';
             }
           } catch (err) {
             // The tag does not exist so let's create it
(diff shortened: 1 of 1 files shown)


nhuet:
Note: this PR assumes that #139 is merged to actually work, as it is using postBuild file.

galleon:
@nhuet I don't think we need this. See #146

nhuet:
Actually this is needed independently from #146 if we want nightly build on default binder (which we have to be sure). I develop:
- #146 is indeed needed if we go the way #139 and #140, 
- if we do all that, it is to get the last nightly build each time we launch default binder
- but when launching this env (by going to https://mybinder.org/v2/gh/airbus/scikit-decide/binder), it will pull an existing already built docker image if no commit has been done since last time someone get there. And it will be the case after the first built following the last release. The postBuild scrit is not executed each time the environment is called but only when the docker image is built. It is called postBuild but it is more lastScriptLaunchWhenBuildingTheEnvironment :)

galleon:
Indeed `postBuild` is misleading.
We then have to  use `start` to install the latest nightly when the container is started (as an ENTRYPOINT)
By triggering using a SHA for past versions or binder for nightly:
  - `environment.yml will always install either the requested version or nothing
  - `start` will do what `postBuild` is currently doing

nhuet:
Great! I totally missed this file. If we want to do that (move from postBuild to start) we indeed need to install all needed dependencies in environment.yml so that start will be very fast (downloading the wheel and installing it without downloading anything else). We have to check that this start remains short (within 10'' i would say) so that it does not penalize the user.

nhuet:
I close for now as this PR needed something like PR #139 that i also closed.

PR #145, opened 2021-12-08
Remove token for github-script

galleon:
[All is in the title]


dbarbier (changes requested):
Please explain in the commit message why this change is needed

dbarbier (approved):

PR #147, opened 2021-12-09
Remove log of non existing  variable

galleon:
The `doIProceed`variable is coming from a former codebase and was not removed everywhere causing an error when run.


dbarbier (approved):
You could also remove most of these logs, less code less bugs.

nhuet (approved):

PR #148, opened 2021-12-09
Binder

galleon:
@nhuet my binder for comments!

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,6 +3,11 @@ dependencies:
   - python=3.8
   - pip
   - pip:
+    - matplotlib
     - ipywidgets
+    - ipympl
     - scikit-decide[all]==0.9.2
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -1,13 +1,10 @@
...
+name: nb-skdecide-py38
 dependencies:
   - python=3.8
   - pip
   - pip:
-    - matplotlib
     - ipywidgets
     - ipympl
-    - scikit-decide[all]==0.9.2
-    - nbgitpuller
```
nhuet, line 10: We do not want to remove nbgitpuller. This is the package used to pull the notebooks from appropriate branch when launching jupyter on binder. (we do neither want to remove matlplotlib as it is necessary and we want to keep as many dependencies in the environment as possible to avoid installing too much at starting time)

PR #150, opened 2021-12-09
use parallel workflows

galleon:
This PR structures the build in 3 different branches:

- each branch is in charge of an operating system family

- each build/test job takes 2 parameters to define a strategy : 
    - one for the python versions
 
    -  one for the different versions of the os we want to build or test on
        - we currently build onto the `latest` OS except for windows were we use `windows-2016`. Latest versions are currently: 
        - we test on more versions. Currenty:
            - windows-2016
            - macos-10.15 and macos-11
            - unbuntu-18.04 and ubuntu-20.04

- the doc is build as soon as possible i.e. as soon as the tests on linux are finished.

This structure does not increase the build time (it slightly decreases it

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,31 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  BOOST_DIR: 3rdparty/boost
+  BOOST_VERSION: "1.76.0"
+  SKDECIDE_SKIP_DEPS: 1
+
 jobs:
+
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      python_version: ${{ steps.generate-matrix.outputs.python_version }}
+      build: ${{ steps.generate-matrix.outputs.build}}
+      test: ${{ steps.generate-matrix.outputs.test}}
+    steps:
+      - uses: actions/setup-python@v2
+      - name: Generate Matrix
+        id: generate-matrix
+        shell: python3 {0}
+        run: |
+          python_version = ["3.7", "3.8", "3.9"]
+          build = { "macos": ["macos-10.15"], "linux": ["ubuntu-latest"], "windows": ["windows-2016"] }
+          test = { "macos": ["macos-latest"], "linux": ["ubuntu-latest"], "windows": ["windows-latest"] }
+          print(f"::set-output name=python_version::{python_version}")
+          print(f"::set-output name=build::{build}")
+          print(f"::set-output name=test::{test}")
 
   lint-sources:
     runs-on: ubuntu-latest
@@ -24,21 +48,18 @@ jobs:
           python-version: "3.8"
       - uses: pre-commit/action@v2.0.0
 
-  build:
+  build-windows:
+    needs:
+      - setup
     strategy:
-      fail-fast: false
       matrix:
-        os: [windows-2016, macos-10.15, ubuntu-latest]
-        python-version: ["3.7", "3.8", "3.9"]
+        os: ${{ fromJSON(needs.setup.outputs.build).windows }}
+        python-version: ${{ fromJSON(needs.setup.outputs.python_version) }}
+      fail-fast: false
     defaults:
       run:
         shell: bash
     runs-on: ${{ matrix.os }}
-
-    env:
-      BOOST_DIR: 3rdparty/boost
-      BOOST_VERSION: "1.76.0"
-      SKDECIDE_SKIP_DEPS: 1
 
     steps:
       - name: Checkout scikit-decide source code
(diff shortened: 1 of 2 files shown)


dbarbier:
How are reviewers supposed to review your changes if you do not explain what these changes are supposed to do?

galleon:
This build workflow:
  - split the build job in three jobs one per os
  - each build/test job takes 2 parameters to define a strategy : one for the python version and one for the versions of the os to build/test on
  - the doc is build as soon as possible (i.e. as soon as tests on linux finish)

.github/workflows/build.yml
```diff
@@ -15,6 +15,37 @@ concurrency:
...
+        id: generate-matrix
+        run: |
+          echo "Set Matrix"
+          echo "::set-output name=python_versions::['3.7', '3.8', '3.9']"
+          echo "::set-output name=build_windows::['windows-2016']"
+          echo "::set-output name=test_windows::['windows-latest']"
+          echo "::set-output name=build_macos::['macos-latest']"
+          echo "::set-output name=test_macos::['macos-10.15', 'macos-11']"
+          echo "::set-output name=build_linux::['ubuntu-latest']"
+          echo "::set-output name=test_linux::['ubuntu-18.04', 'ubuntu-20.04']"
```
dbarbier, line 39: We had previously:
 * Build matrix:  [windows-2016, macos-10.15, ubuntu-latest]
 * Test matrix:  [windows-latest, macos-latest, ubuntu-latest]

This PR should mimic that, and configuration changes are made in another PR.

.github/workflows/build.yml
```diff
@@ -15,6 +15,37 @@ concurrency:
...
+
+  my_echo:
```
dbarbier, line 41: Please remove this job; do not keep debugging stuff in production code, this is useless and may add more bugs

.github/workflows/build.yml
```diff
@@ -24,12 +55,98 @@ jobs:
...
+      matrix:
+        os: ${{ fromJSON(needs.init.outputs.build_windows) }}
+        python-version: ${{ fromJSON(needs.init.outputs.python_versions) }}
       fail-fast: false
+    defaults:
+      run:
+        shell: bash
+    runs-on: ${{ matrix.os }}
+
+    env:
```
dbarbier, line 71: env is the same for all jobs and could be declared globally

.github/workflows/build.yml
```diff
@@ -24,12 +55,98 @@ jobs:
...
+            skdecide/hub/bin
+            skdecide/hub/share
+            skdecide/hub/*.msc
+          key: ${{ runner.os }}-cache-deps-${{ hashFiles('cpp/deps/chuffed', 'cpp/deps/gecode', 'cpp/deps/libminizinc') }}
+
+      - name: Update SKDECIDE_SKIP_DEPS
+        if: steps.cache-build-dependencies.outputs.cache-hit != 'true'
+        run: echo "SKDECIDE_SKIP_DEPS=0" >> $GITHUB_ENV
+
+      - name: Build Windows wheel
```
dbarbier, line 125: Remove mentions of OS from step names (in all steps), this is no more useful when jobs are split by OS.

.github/workflows/build.yml
```diff
@@ -160,18 +347,26 @@ jobs:
...
-        os: [ubuntu-latest]
-        python-version: ["3.7", "3.8", "3.9"]
+        # os: ${{ fromJSON(needs.init.outputs.test_windows) }}
+        os: [windows-latest]
+        python-version: ${{ fromJSON(needs.init.outputs.python_versions) }}
         compiler: [gnu]
       fail-fast: true
+    defaults:
+      run:
+        shell: bash
```
dbarbier, line 363: No idea why, but test failures occur only when running in bash, remove that stance to have the same behavior as before.  Remove it from all 3 test-* jobs, they were not set previously and are useless anyway on Unix.

.github/workflows/build.yml
```diff
@@ -352,7 +553,7 @@ jobs:
 
   upload-nightly:
     if: (github.ref == 'refs/heads/master') && (github.repository == 'airbus/scikit-decide') && (github.event_name == 'schedule')
-    needs: [test-unix, test-macos, test-windows]
+    needs: [build-docs, test-macos, test-windows]
```
dbarbier, line 556: why this change?

dbarbier (changes requested):

.github/workflows/build.yml
```diff
@@ -24,12 +55,98 @@ jobs:
...
       - uses: pre-commit/action@v2.0.0
 
-  build:
+  build-windows:
+    needs:
+      - init
     strategy:
+      matrix:
+        os: ${{ fromJSON(needs.init.outputs.build_windows) }}
+        python-version: ${{ fromJSON(needs.init.outputs.python_versions) }}
```
dbarbier, line 64: python-version on the left and python_version**s** on the right, that is not ideal

.github/workflows/build.yml
galleon: As explained in the description of the PR. Building the doc is done as early as possible. Since it is done on `ubuntu` it is done after tests have been performed. Therefore `build-docs` depends now from `test-linux` and `upload-nightly` depends on `build-docs`

.github/workflows/build.yml
galleon: ok back to previous state.

.github/workflows/build.yml
galleon: done - back to previous non // set-up

.github/workflows/build.yml
galleon: done

.github/workflows/build.yml
galleon: done

.github/workflows/build.yml
galleon: fixed

.github/workflows/build.yml
dbarbier: But upload-nightly has nothing to do with docs, so why does it depend to build-docs?

.github/workflows/build.yml
dbarbier:
```sh
$ grep -i -E 'name:.*(linux|macos|windows)' .github/workflows/build.yml
      - name: Install omp on MacOS
      - name: Install and restore ccache for macOs
      - name: Let cmake use ccache for macOs
      - name: Restore docker dev image for Linux
      - name: Restore ccache cache for Linux
```

galleon: > grep -i -E 'name:.*(linux|macos|windows)' .github/workflows/build.yml                                                [🐍 scikit-decide]
>

.github/workflows/build.yml
dbarbier: BTW it will be simpler to write this step in Python when you will make this workflow more complex; db/shell-python does that.  (And db/shell-python-extended shows how this Python code can be modified in order to implement some complex processing in future PRs, if desired)

galleon: done - taken

.github/workflows/build.yml
galleon: done

.github/workflows/build.yml
galleon: fixed

.github/workflows/build.yml
galleon: ok went back to same structure

.github/workflows/build.yml
```diff
@@ -13,8 +13,32 @@ concurrency:
...
+    steps:
+      - uses: actions/setup-python@v2
+      - name: Generate Matrix
+        id: generate-matrix
+        shell: python3 {0}
+        run: |
+          python_version = ["3.7", "3.8", "3.9"]
+          build = { "macos": ["macos-10.15"], "linux": ["ubuntu-latest"], "windows": ["windows-2016"] }
+          test = { "macos": ["macos-latest"], "linux": ["ubuntu-latest"], "windows": ["windows-latest"] }
+          print("::set-output name=python_version::['3.7', '3.8', '3.9']")
```
dbarbier, line 38: You should write
```python
print(f"::set-output name=python_version::{python_version}")
```

.github/workflows/release.yml
```diff
@@ -5,22 +5,132 @@ on:
...
+    steps:
+      - uses: actions/setup-python@v2
+      - name: Generate Matrix
+        id: generate-matrix
+        shell: python3 {0}
+        run: |
+          python_version = ["3.7", "3.8", "3.9"]
+          build = { "macos": ["macos-10.15"], "linux": ["ubuntu-latest"], "windows": ["windows-2016"] }
+          test = { "macos": ["macos-latest"], "linux": ["ubuntu-latest"], "windows": ["windows-latest"] }
+          print("::set-output name=python_version::['3.7', '3.8', '3.9']")
```
dbarbier, line 30: `print(f"::set-output name=python_version::{python_version}")`

dbarbier (changes requested):
build.yml looks good, you can squash commits after changing the setup step as suggested.

But release.yml is wrong, you copy/pasted build.yml, you should instead have duplicated the steps in release.yml.  We do not want to use ccache in release.yml, maybe there are other differences, I do not want to check, this would require too much time.

In fact release.yml could be left as is, we won't implement [ci: skip] keywords in this workflow, so it is less important.  If you do want to change it anyway, please make sure that the new release.yml does the exact same thing as the old one.

dbarbier:
[...]
> In fact release.yml could be left as is, we won't implement [ci: skip] keywords in this workflow, so it is less important. If you do want to change it anyway, please make sure that the new release.yml does the exact same thing as the old one.

And if you want to modify release.yml, you should rebase on current master which also modified this file.

galleon:
I have re-squashed and kept 2 commits one for each file. 
Applying the same pattern on release should allow later to skip pushing release to PyPI if needed or doing only documentation publication.

.github/workflows/build.yml
dbarbier: still not fixed after rebase

dbarbier:
> Applying the same pattern on release should allow later to skip pushing release to PyPI if needed

I cannot imagine why that would be useful.

> or doing only documentation publication.

Since release.yml is only triggered by pushing tags, this would mean that you delete and re-push a tag for a published version, this is certainly not desirable.

.github/workflows/build.yml
galleon: fixed.

.github/workflows/release.yml
galleon: done

dbarbier (approved):

.github/workflows/build.yml
dbarbier: still not fixed

dbarbier (changes requested):

galleon:
Indeed the changes for `release.yml` are targetting forks only (i.e. tests before tagging on main repository).

dbarbier (approved):

PR #151, opened 2021-12-10
Skip windows build

galleon:
THis PR build on the parallel workflow one.

This PR will skip the builds and tests on windows if the string "ci: skip-windows" appears in the commit message


galleon:
Will be resubmitted later - or never

PR #156, opened 2021-12-13
Trigger a build of release binder env on mybinder.org

nhuet:
The goal is to reduce the starting time for real users. Indeed,
according to the mybinder documentation, the build step occurs only the
first time the binderhub launches a jupyter session for a repo with an updated ref. Cf https://mybinder.readthedocs.io/en/latest/about/about.html#what-factors-influence-how-long-it-takes-a-binder-session-to-start

We use a script taken from https://github.com/scikit-hep/pyhf/blob/master/binder/trigger_binder.sh
to ping the endpoints of each binderhub in the mybinder.org federation.
(cf https://mybinder.readthedocs.io/en/latest/about/federation.html)

We do not need to stay connected to the binderhub for the build to go on, so the trigger step in the workflow takes less than 1' as one can see in the workflow generated by a fake release on my fork: https://github.com/nhuet/scikit-decide/runs/4505623076?check_suite_focus=true 
We can see that the build has been done for the commit corresponding to the updated binder environment:
https://mybinder.org/v2/gh/nhuet/scikit-decide/a0e09be06f993f0000b4d6c4a2908f7eec3e266f

(Here you have to trust me for not having follow this link before the workflow took place :)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -505,6 +505,13 @@ jobs:
           # switch back to original branch
           git checkout $TAG_NAME
 
+      - name: Trigger a build on each BinderHub deployments in the mybinder.org federation
+        run: |
+          bash scripts/trigger_binder.sh https://gke.mybinder.org/build/gh/nhuet/scikit-decide/${BINDER_RELEASE_ENV_SHA1}
+          bash scripts/trigger_binder.sh https://ovh.mybinder.org/build/gh/nhuet/scikit-decide/${BINDER_RELEASE_ENV_SHA1}
+          bash scripts/trigger_binder.sh https://turing.mybinder.org/build/gh/nhuet/scikit-decide/${BINDER_RELEASE_ENV_SHA1}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -505,6 +505,13 @@ jobs:
           # switch back to original branch
           git checkout $TAG_NAME
 
+      - name: Trigger a build on each BinderHub deployments in the mybinder.org federation
+        run: |
+          bash scripts/trigger_binder.sh https://gke.mybinder.org/build/gh/nhuet/scikit-decide/${BINDER_RELEASE_ENV_SHA1}
```
dbarbier, line 510: should not be nhuet there

dbarbier (changes requested):

.github/workflows/release.yml
nhuet: :scream:

nhuet:
Retesting release workflow after correction of repository name in https://github.com/nhuet/scikit-decide/actions/runs/1576990657

dbarbier (approved):

PR #158, opened 2021-12-13
Fix #133 (Docs is not building with node 16)

### LINKED ISSUE #133: Doc is not building with node 16
## 🐛 Bug

GH workers are now using node 16 as their default node environment. This causes the build of the doc to break

## To Reproduce

1. Remove the node set-up in build / release workflow

nhuet:
Call the function updateLinks() which references document (ie DOM)
inside a `mounted` hook.

Indeed the previous code was working with `yarn vuepress dev docs` but
not with `yarn vuepress build docs` because when *building*, vue uses a
server to generate the static website and usually `document` is only
available on client side.

However this is possible when inside the mounted hook
(https://vuepress.vuejs.org/guide/using-vue.html#browser-api-access-restrictions)

NB: no idea why this was working with node 14 though...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -486,10 +486,6 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - uses: actions/setup-node@v2
-        with:
-          node-version: '14'
-
       - name: Download artifacts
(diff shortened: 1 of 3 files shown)


nhuet:
Fix #133

PR #161, opened 2021-12-14
Modify titles of domain/solver specification boxes

nhuet:
Following a suggestion of @neo-alex , the boxes should not have "TIP" and "WARNING" titles but rather "Domain specification" and "Solver specification". (The type of boxes are only here to format highlight them)

Previously:

Screenshot from 2021-12-14 12-20-54

New version:

Screenshot from 2021-12-14 12-21-41

One can check the whole new generated website with the following process:
- downloading the "doc" artifact from the "Build scikit-decide" action triggered by this PR 
- unzip and serving it with
 ```shell
  mkdir testdoc
  cd testdoc
  unzip path/to/doc.zip -d scikit-decide
  python -m http.server
  ```


nhuet:
Following a suggestion of @neo-alex

PR #162, opened 2021-12-14
Provide filters for ci builds

galleon:
By using specific keywords the user can interact on the ci build and release workflows
For `build` the following keywords are available:
- By default, built are done on the oldest versions of the oses
- [ci: os-version] will keep the specific version of the  given os (if version is omitted all versions will be kept),
- [skip-doc] will skip the doc generation

For `release`, only one keyword is available:
- [ci: skip-deploy-test-pypi] will skip the deployment of the wheels on Test PyPI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,6 +26,10 @@ jobs:
       python_version: ${{ steps.generate-matrix.outputs.python_version }}
       build: ${{ steps.generate-matrix.outputs.build}}
       test: ${{ steps.generate-matrix.outputs.test}}
+      do_macos: ${{ steps.generate-matrix.outputs.do_macos}}
+      do_ubuntu: ${{ steps.generate-matrix.outputs.do_ubuntu}}
+      do_windows: ${{ steps.generate-matrix.outputs.do_windows}}
+      build_doc: ${{ steps.generate-matrix.outputs.build_doc}}
     steps:
       - uses: actions/setup-python@v2
       - name: Generate Matrix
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -26,18 +26,63 @@ jobs:
...
       - uses: actions/setup-python@v2
       - name: Generate Matrix
         id: generate-matrix
         shell: python3 {0}
         run: |
+          # Default variables
           python_version = ["3.7", "3.8", "3.9"]
-          build = { "macos": ["macos-10.15"], "linux": ["ubuntu-latest"], "windows": ["windows-2016"] }
-          test = { "macos": ["macos-10.15", "macos-11"], "linux": ["ubuntu-18.04", "ubuntu-20.04"], "windows": ["windows-2016", "windows-2019", "windows-2022"] }
+          build = [ "macos-10.15", "ubuntu-18.04", "windows-2016" ]
```
dbarbier, line 41: I fail to see how changing ubuntu-latest into ubuntu-18.04 is related to this PR.  Anyway this change is useless.

.github/workflows/build.yml
```diff
@@ -443,8 +491,12 @@ jobs:
           pytest -v -s tests/solvers/python
           pytest -v -s tests/scheduling
 
-  build-docs:
-    needs: [test-linux, test-macos, test-windows]
+  build-doc:
+    needs:
+      - test-ubuntu
+      - test-macos
+      - test-windows
```
dbarbier, line 498: Why do you suddenly decide that 'needs' should be written on multiple lines?  Previous version looks more readable to me

.github/workflows/build.yml
```diff
@@ -26,18 +26,63 @@ jobs:
...
       - name: Generate Matrix
         id: generate-matrix
         shell: python3 {0}
         run: |
+          # Default variables
           python_version = ["3.7", "3.8", "3.9"]
-          build = { "macos": ["macos-10.15"], "linux": ["ubuntu-latest"], "windows": ["windows-2016"] }
-          test = { "macos": ["macos-10.15", "macos-11"], "linux": ["ubuntu-18.04", "ubuntu-20.04"], "windows": ["windows-2016", "windows-2019", "windows-2022"] }
+          build = [ "macos-10.15", "ubuntu-18.04", "windows-2016" ]
+          test = [ "macos-10.15", "macos-11", "ubuntu-18.04", "ubuntu-20.04", "windows-2016", "windows-2019" ]
```
dbarbier, line 42: why did you remove windows-2022?  This is not mentioned in your description
BTW it builds without error.

.github/workflows/release.yml
```diff
@@ -18,18 +18,67 @@ jobs:
...
+              python_version = [v for v in python_version if python_filter[v]]
+          os_filter = {
+              'macos-10.15'  : to_bool("${{ contains(github.event.head_commit.message, '[ci: macos-10.15]') }}"),
+              'macos-11'     : to_bool("${{ contains(github.event.head_commit.message, '[ci: macos-11]') }}"),
+              'ubuntu-18.04' : to_bool("${{ contains(github.event.head_commit.message, '[ci: ubuntu-18.04]') }}"),
+              'ubuntu-20.04' : to_bool("${{ contains(github.event.head_commit.message, '[ci: ubuntu-20.04]') }}"),
+              'windows-2016' : to_bool("${{ contains(github.event.head_commit.message, '[ci: windows-2016]') }}"),
+              'windows-2019' : to_bool("${{ contains(github.event.head_commit.message, '[ci: windows-2019]') }}"),
+          }
+          if set(os_filter.keys()) != set(test):
```
dbarbier, line 55: Your description states that there is only one keyword parsed by release.yml.  This is obviously false.

dbarbier (changes requested):

.github/workflows/build.yml
galleon: This is the implementation of:
"By default, built are done on the oldest versions of the oses"

If you build on the latest os, you cannot guarantee that it will work on an oldest one.

.github/workflows/release.yml
galleon: The statement means that only one keyword is active in `release`. Ideally the setup could/should be the same for both workflow.

.github/workflows/build.yml
dbarbier: Yes we can™, this is why we build with manylinux on Linux

.github/workflows/release.yml
dbarbier: > The statement means that only one keyword is active in release.

Which is false.  It would be much more obvious if you did not copy everything but only keep the keyword you want to parse.

dbarbier: Okay I stand corrected, my statement is wrong, all keywards are active, but needs.setup is never used. But this is error prone, dead code should be removed.

.github/workflows/build.yml
```diff
@@ -443,8 +483,9 @@ jobs:
           pytest -v -s tests/solvers/python
           pytest -v -s tests/scheduling
 
-  build-docs:
-    needs: [test-linux, test-macos, test-windows]
+  build-doc:
+    needs: [test-ubuntu, test-macos, test-windows]
```
dbarbier, line 487: you must add setup in needs, docs are currently not built.

galleon:
Release tested and results visible here: https://github.com/galleon/scikit-decide/actions/runs/1583655111

dbarbier (approved):

PR #164, opened 2021-12-14
Remove explicit dependency in torch.

nhuet:
Torch will still be installed via stable_baselines3 when [solver] or [all] is selected, but the version will not be specified by scikit-decide itself.


galleon (approved):
looks good to me

PR #166, opened 2021-12-16
Fix missing parenthesis in notebooks commented code

nhuet:
In a comment, we show how to solve the domain inside
a `with` block. The code was broken because of a missing ).


galleon (approved):
Looks good to me

PR #168, opened 2021-12-16
Removing need for GH_ACCESS_TOKEN

galleon:
The token was not in the secrets which causes this job to break.
The secret is not referenced anymore

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -551,7 +551,6 @@ jobs:
       - name: Delete nightly release
         uses: actions/github-script@v5
         with:
-          github-token: ${{ secrets.gh_access_token }}
           script: |
             const releases = await github.rest.repos.listReleases({
               owner: context.repo.owner,


nhuet (commented):
I am pretty sure it is working, but did you test it (=the new release wf) on a tag on your fork? Can you add a link to the corresponding run if it is the case? 
Thanks !

(If i were picky, i would point out that we would rather have first line of commit line be in imperative mood  :)

galleon:
It was tested live: https://github.com/airbus/scikit-decide/runs/4551726615?check_suite_focus=true. On my repo I have a secret :-) but it is not needed ... It was removed everywhere but this one was forgotten.

dbarbier (approved):

nhuet:
Hum... When following your link, i see a workflow crashing :sweat_smile: : 
> Error: Unhandled error: Error: Input required and not supplied: github-token

So it seems not working actually ?

nhuet:
Ok, i understand now, the link was pointing to the airbus repo (and not @galleon 's fork) and to the release occuring before this PR. So the link was explaining why the PR is needed. 
Personnally i would have tested with a fake release on my fork with this commit to be sure, but it is ok for now.

PR #169, opened 2021-12-17
Test python code examples shown in guide during CI

### LINKED ISSUE #165: Test the code in the guide and other pages
## 📚 Documentation

The code in the section "How to" of online guide is broken. A good idea could be to use libraries like [doctest](https://docs.python.org/3/library/doctest.html) to test these portions of code as part of the CI process, so that broken code is seen as soon at it happens.

nhuet:
Test python code examples shown in guide
    
- script/md2py.py create a python script from the python code blocks found in a markdown file
- it uses docopt (to parse arguments) and commonmark (to parse markdown)
- we run the resulting python script in CI

NB: if we launch the script with pytest instead of python, it is passing (pytest does not complain) but the exit code of pytest is 5 (no actual tests found) so that github action thinks that the test fails.

Fix #165.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -467,7 +467,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install pytest tqdm
+          python -m pip install pytest tqdm docopt commonmark
 
       - name: Download artifacts
         uses: actions/download-artifact@v1.0.0
@@ -481,6 +481,11 @@ jobs:
           pytest -v -s tests/solvers/cpp
           pytest -v -s tests/solvers/python
           pytest -v -s tests/scheduling
+
+      - name: Test python block codes from guide
+        run: |
+          python scripts/md2py.py docs/guide/README.md tests/test_guide.py
+          python tests/test_guide.py
 
   build-doc:
     needs: [test-ubuntu, setup]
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -467,16 +467,21 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install pytest tqdm
+          python -m pip install pytest tqdm docopt commonmark
+          pip install --pre --find-links ./wheels/ "scikit-decide[all]"
```
dbarbier, line 471: you are testing the last released skcikit-decide, not the one just being built.  Move your new test after pytest

dbarbier (changes requested):

.github/workflows/build.yml
nhuet: indeed...

nhuet: I will squash my corrections when you accept them.

dbarbier: Please make changes as minimal as possible, there is no reason to move the "Install dependencies" stanza

nhuet: Ok. 
I just do not understand why scikit-decide is installed during "test" step rather than in "install" step. But indeed my changes are not minimal, I agree.
And i do understand why minimal changes are better.
I force pushed a single commit so that only minimal changes are displayed.

dbarbier: >  And i do understand why minimal changes are better.

They are easier to review

nhuet: Yes exactly

dbarbier (approved):

PR #170, opened 2021-12-18
Fix environment.yml update in release.yml

dbarbier:
This workflow pushes a commit which modifies environment.yml to require
the just released version, then reverts this commit.  But there is no
reason to revert this commit.


nhuet (approved):
I agree. 

The revert will be needed when the environment defined by binder branch do install the dev version of scikit-decide, but this is not yet the case.

PR #172, opened 2021-12-18
Install torch with cpuonly from pytorch channel

dbarbier:
This version is 200MB instead of 900MB, and binders do not have GPUs.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,7 +1,12 @@
 name: nb-skdecide092-py38
+channels:
+  - pytorch
+  - defaults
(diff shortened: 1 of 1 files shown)


nhuet (approved):
Much readable via conda than via pip indeed. 

(By the way, i found it was possible via pip requirements without postBuild with 
```
    - torch==1.10.0+cpu --install-option="--find-links https://download.pytorch.org/whl/cpu/torch_stable.html" 
``` 
but i prefer also the conda version.)

Issue #142, opened 2021-12-08
scikit does not compile with Visual Studio 16 2019 


galleon:
## 🐛 Bug

When.moving to `windows-latest` on GitHubActions we got the following environment:
-- Building for: Visual Studio 16 2019
-- The C compiler identification is MSVC 19.29.30137.0
-- The CXX compiler identification is MSVC 19.29.30137.0

This causes the following error:

```
D:/a/scikit-decide/scikit-decide/cpp/src/hub/solver/riw/impl/riw_impl.hh(585,5): message : existing declarations [D:\a\scikit-decide\scikit-decide\build\temp.win-amd64-3.7\Release_skdecide\hub\__skdecide_hub_cpp\src\hub\solver\riw\py_riw.vcxproj]
[...]

PR #173, opened 2022-01-11
Fix issue 142

fteicht:
Fix issue #142


cpp/src/hub/solver/mcts/mcts.hh
```diff
@@ -97,17 +97,11 @@ public:
       HeuristicFunctor;
 
   FullExpand(
-      const HeuristicFunctor &heuristic =
+      const HeuristicFunctor &heuristic =
```
dbarbier, line 100: whitespace changes are bad (and break linters by the way)

cpp/CMakeLists.txt
```diff
@@ -18,6 +18,10 @@ IF (NOT WIN32 AND NOT CYGWIN)
     SET(CMAKE_CXX_FLAGS "-fvisibility=hidden -fvisibility-inlines-hidden ${CMAKE_CXX_FLAGS}")
 ENDIF ()
 
+IF (MSVC)
+    ADD_COMPILE_OPTIONS(/MP /Zc:lambda)
+ENDIF ()
```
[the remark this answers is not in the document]
dbarbier: Hmmm it looks like the 3rd item was wrong, MSVC silently ignores unknown flags.  But this is not true for clang, so my recommendation still stands.

dbarbier:
@fteicht Your changes break windows-2016 builds; any chance to have a version which works with any windows-* version?

cpp/CMakeLists.txt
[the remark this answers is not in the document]
fteicht: Actually older versions of MSVC silently ignore the lambda compilation flag but they cannot parse the simplified version of MCTS' FullExpand's constructor (another independent issue). So I will revert the corresponding commit.

fteicht:
> 

Yes because I've just figured out that old versions of MSVC (e.g. 2017 and before) cannot correctly parse typedef members inside lambda functions. It was corrected by the introduction of the `Zc:lambda` compilation flag but it only works with newest versions of MSVC. Thus I've reverted the commit that simplifies MCTS' FullExpand's constructor implementation.

cpp/src/hub/solver/riw/riw.hh
```diff
@@ -215,6 +215,13 @@ private:
     typedef Trollout_policy<Domain> RolloutPolicy;
     typedef Texecution_policy ExecutionPolicy;
 
+  private:
+    typedef typename ExecutionPolicy::template atomic<std::size_t>
+        atomic_size_t;
+    typedef typename ExecutionPolicy::template atomic<double> atomic_double;
+    typedef typename ExecutionPolicy::template atomic<bool> atomic_bool;
```
dbarbier, line 222: These typedefs appear in public methods, why are they private?

dbarbier (approved):
Well done, looks good to me.

Since windows-2016 Github runners are going to be removed on March 2022, you should replace windows-2016 by windows-latest on l.40 of build.yml 
          build = [ "macos-10.15", "ubuntu-latest", "windows-2016" ]
and l. 29 of release.yml
          build_dict = { "macos":["macos-10.15"], "ubuntu":["ubuntu-latest"], "windows":["windows-2016"] }
On Windows backward compatibility is ensured, so we can build on -latest.

And bonus points if you squash your commits ;-)

cpp/src/hub/solver/riw/riw.hh
fteicht: You're right, I've just committed the change. Thanks!

dbarbier:
@fteicht  Please remove the merge commit, it does not provide any value:

```
$ git diff 45a5132 7bf5f70
$
```

You can run
```
 git reset --hard 45a5132
 git push --force-with-lease origin fix-issue-142
 ```

PR #174, opened 2022-01-11
Optimisation of scheduling domains efficiency

g-poveda:
Most work rely on https://github.com/dbarbier/scikit-decide/tree/profiling-scheduling branch.
Execution of a scheduling policy is currently possible in two modes "inplace=True" or "inplace=False". The inplace=True is modifying the scheduling state without doing a copy of it, therefore we can't "backward" easily and the use of many kind of algorithm that involve internal graph building is impossible.
We observed that the option "inplace=False" (which creates a brand new object at every simulation step) lead to a computation time up to 20x slower than the other version.

Modelling state changes had been explored notably the use of singly linked list containing all the infos of completed task (therefore frozen for the rest of the execution). It proved to reduce the overhead of the inplace=False option.

Few adaptation of codes had been done so that the examples and unit test work with this new modelling :  
- plots function was reading ```task_details``` attribute of state object
- unit test checks also heavily relied on the same attribute
- gphh algorithm was running some schedule generation scheme that was taking into account the task_details.
The basic fix of all this problems are straightforward, by using a function that rebuild the same information as before, from the singly linked list.


notebooks/profiling-scheduling.ipynb
```diff
@@ -0,0 +1,366 @@
+{
```
galleon, line 1: If the notebook is added in that directory it might be worth to follow the right naming scheme.
Can you also provide binder to check how it renders.

galleon (changes requested):

notebooks/profiling-scheduling.ipynb
g-poveda: we have to decide if the notebook and the examples/scheduling/computation_time_experiments.py should be in the PR or not. @dbarbier could have views on this.

dbarbier: My branch was not meant to be merged, individual commits should be cherry-picked and polished.
For instance we saw during our meeting that commit 3467728 was wrong, as mentioned in the log message, the test is buggy and must be fixed instead.   We were  also quite reluctant to merge 5239abf.

g-poveda: Ok I did the cherry picking finally and will do another PR because I pushed in one another branch in my repo, so it is ok that I close this one and start a new ? thanks

dbarbier: Sure, thanks.

g-poveda:
Reopened a cleaner PR for the same main topic : https://github.com/airbus/scikit-decide/pull/185

PR #176, opened 2022-01-13
Fix issue 175

galleon:
This commit change the way a `Box` action is discretized.
It is inspired from the `gym.spaces.Box.sample` method. For `unbounded` spaces it builds on the quantile distribution of the Cauchy distribution (using a scale of 1).
An alternative could be derived from the `logit` function: `log(1+p) - log(1-p)` for example. But the current implementation has a larger `spread`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -551,7 +551,6 @@ jobs:
       - name: Delete nightly release
         uses: actions/github-script@v5
         with:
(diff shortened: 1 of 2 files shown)


nhuet:
Hi. The first commit "[Removing need for GH_ACCESS_TOKEN" should not be part of this PR. For the remaining, it seems ok for me but I let more qualified people decide about what the Box space should look like.

PR #185, opened 2022-04-07
Optimize scheduling domains code

nhuet:
We use an example (added in first commit) to try to improve scheduling domain code.

The main idea is to reduce time passed in copying the state,
by modifying the way the information on tasks is stored.

More precisely, this is the list of modifications done:

- Remove resources (unused) from tasks_details
- Lazy initialization of tasks_details
- Initialize start and sample_duration in Task constructor
- Store in state information on completed tasks in separate attributes as
  singly linked lists, since they will not be changed anymore:
  tasks_complete_details, tasks_complete_progress and tasks_complete_mode.
  Indeed singly linked lists can avoid duplication of tasks_details,
  tasks_progress and tasks_mode:
    * when a Task is inserted, it becomes the new head node and points to
      the previous head
    * in State.copy, a new singly linkedlist is created, but with the
      same head as previous state.  Both lists have the same nodes, but
      nodes inserted after can be different.

  The only drawback is that lookup is now O(N)

- Remove State.tasks_remaining.
  This set is heavy and makes State.copy() slow.
  State contains the following attributes (among others):
    * task_ids: a list of all task ids in the scheduling domain
    * tasks_remaining: a set containing the ids of tasks still to be started
    * tasks_complete: a set containing the ids of tasks that have been
          completed
    * tasks_ongoing: a set containing the ids of tasks started and not
          paused and still to be completed
    * tasks_paused: a set containing the ids of tasks that have been started
          and paused but not resumed yet
    * tasks_progress: a dictionary where the key is a task id (int) and
          the value the progress of the task between 0 and 1 (float)
  Remark: keys(tasks_progress) = tasks_ongoing + tasks_paused

  The idea behind tasks_remaining is that some tasks may require
  conditions which are not fulfilled, so tasks_remaining is a subset of
  task_ids which contains tasks which can be started.

  But at the beginning of simulation, this set may be large, and is copied
  into all State instances.  We replace this member by
    * tasks_unsatisfiable: a set containing the ids of tasks for which
          conditions are not fulfilled

  We can use tasks_unsatisfiable to compute tasks_remaining by:

     task_ids - tasks_complete - tasks_progress - tasks_unsatisfiable

  Copy of tasks_unsatisfiable will be much faster than tasks_remaining,
  but the extra computations may cancel this gain, the overall speedup
  (if any) must be confirmed on real cases.

- Speedup MixedRenewable.all_tasks_possible() buy moving creation of
  dicts and sets outside of loops.
- Do not call get_task_modes() in loops

### DIFF
--- a/skdecide/builders/domain/scheduling/scheduling_domains_modelling.py
+++ b/skdecide/builders/domain/scheduling/scheduling_domains_modelling.py
@@ -4,11 +4,70 @@
 
 from __future__ import annotations
 
-from copy import deepcopy
+from collections.abc import Collection
+from copy import copy, deepcopy
 from enum import Enum
-from typing import Dict, List, Optional, Set, Union
+from typing import Dict, Iterable, Optional, Set, Tuple, Union
 
 from skdecide.builders.domain.scheduling.task import Task
+
+
+def rebuild_tasks_complete_details_dict(state: State) -> Dict[int, Task]:
+    tasks_complete_details = {p.value.id: p.value for p in state.tasks_complete_details}
+    return tasks_complete_details
+
+
+def rebuild_all_tasks_dict(state: State) -> Dict[int, Task]:
+    tasks_details = {p.value.id: p.value for p in state.tasks_complete_details}
+    tasks_details.update(state.tasks_details)
+    return tasks_details
+
+
+def rebuild_tasks_modes_dict(state: State) -> Dict[int, int]:
+    tasks_modes = {p.value[0]: p.value[1] for p in state.tasks_complete_mode}
+    return tasks_modes
+
+
+def rebuild_schedule_dict(state: State) -> Dict[int, Dict[str, int]]:
+    schedule = {
+        p.value.id: {"start_time": p.value.start, "end_time": p.value.end}
+        for p in state.tasks_complete_details
+    }
+    return schedule
+
+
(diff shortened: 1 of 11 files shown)


skdecide/builders/domain/scheduling/scheduling_domains_modelling.py
```diff
@@ -4,13 +4,72 @@
...
+from collections.abc import Collection
+from copy import copy, deepcopy
 from enum import Enum
-from typing import Dict, List, Optional, Set, Union
+from typing import Dict, Iterable, Optional, Set, Tuple, Union
 
 from skdecide.builders.domain.scheduling.task import Task
 
 
+def rebuild_tasks_complete_details_dict(state: State) -> Dict[int, Task]:
```
g-poveda, line 15: do this "rebuild_***" function deserve some docstring ? I think their name + signature are fine, but you may think differently

g-poveda (commented):
knowing the content of the PR I trust its content and I approve the commit + description. but I will do local small tests before approving merge.

g-poveda:
The changes of API for scheduling states will need to be mentionne in the next release. Otherwise I'm fine with the PR, I let @galleon confirm since he has to review it.

galleon (approved):

PR #188, opened 2022-04-15
Fix minizinc configuration

nhuet:
When importing `skdecide.hub`, the path to minizinc binaries and to
solvers definition is updated with embedded versions of minizinc.

This worked great with install from releases but was not working when
using poetry, because paths are different.
Another issue was that more paths than needed were potentially added.

For release install, it is actually sufficient to look at directory
containing `skdecide.hub.__init__.py` as it should contain the solvers
configuration files *.msc and the bin/ directory with minizinc (and
the C++ library` __skdecide_hub_cpp`)

For poetry, this directory is still needed to find gecode.msc, but the
other solver configurations and the bin/ directory with minizinc are in the
build directory generated by `poetry install`, together with the C++
library  `__skdecide_hub_cpp`, which is used to find the proper directory.

NB:
- in release install case, if `import skdecide.hub` is done when
  working directory is the project directory where a poetry install has
  been done, it can result by pointing to the minizinc built by poetry
  (but anyway it will also be the code from the repository that will be
  used for skdecide instead of the one installed via pip)
- if several poetry installs have been made (with various version of
  python for instance), we do not ensure that the minizinc chosen will
  be that of the current poetry environment.

### DIFF
--- a/docs/contribute.md
+++ b/docs/contribute.md
@@ -75,10 +75,11 @@ Here are the steps to follow:
         poetry env use 3.8.11
         ```
 
-    - Install all dependencies as defined in `poetry.lock`.
+    - Install all dependencies as defined in `poetry.lock`, build c++ libraries, and configure them.
         ```shell
         rm -rf build  # removing previous build
         poetry install --extras all
+        poetry run python postbuild_dev.py
         ```
 
 #### Alternate installation with conda + poetry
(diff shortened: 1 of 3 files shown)


nhuet:
I have tested the changes on unix, but it would be nice to check it also on macos and windows.
To do so, checkout the branch, and
```
poetry install --extras all
poetry run python -c 'from skdecide.discrete_optimization.rcpsp.solver.cp_solvers import Solver; Solver.lookup("gecode"); Solver.lookup("chuffed")'
```
It should crash before this fix and should work now.

nhuet:
I have tested the changes on unix, but it would be nice to check it also on macos and windows.
To do so, checkout the branch, download this [small example script](https://gist.github.com/nhuet/8191d3980bc38c0371bf21641d7fb7c4#file-gecode-py) which uses the minizinc+gecode embedded in scikit-decide, and
```
poetry install --extras all
poetry run python postbuild_dev.py
poetry run python gecode.py
```

PR #189, opened 2022-04-21
Rewrite extender to return a list

nhuet:
We revert commit 9df73f72ebc05c8242e1cfb782e6b96855702af8 because
returning a generator may lead to errors hard to understand when the
domain developper do not take properly into account how "yield" works,
for instance if using global variables.

But we still modify how the list is computed to avoid using 2
intermediate lists (which allow to gain 1.1 speed factor on a
given graph shortest path example).


fteicht (approved):
Perfect, thanks @nhuet !

PR #190, opened 2022-04-25
Fix minizinc configuration

nhuet:
When importing skdecide.hub, the path to minizinc binaries and to
solvers definition is updated with embedded versions of minizinc.

This worked great with install from releases but was not working when
using poetry, because paths are different.
Moreover, more folders than needed were potentially added to PATH
variable.

For release install, it is actually sufficient to look at directory
containing skdecide.hub.__init__.py as it should contain the solvers
configuration files *.msc and the bin/ directory with minizinc (and
the C++ library __skdecide_hub_cpp)

For poetry, except for gecode.msc, the solver configurations and the bin/
directory with minizinc are in the build directory generated by `poetry install`,
together with the C++ library  __skdecide_hub_cpp, which is used to find the proper
directory.
=> skdecide/hub/__init__.py is modified consequently.

For this to work, gecode.msc needs also to be copied to the directory in build/
containing the other solvers configurations like chuffed.msc. Indeed, gecode.msc
"mznlib" field refers to a path relative to it which actually exists in the latter directory.
=> a postbuild_dev.py script is added doing this and the documentation
for developpers is amended to show how to use it.


nhuet:
I have tested the changes on unix, but it would be nice to check it also on macos and windows.
To do so, checkout the branch, download this [small example script](https://gist.github.com/nhuet/8191d3980bc38c0371bf21641d7fb7c4#file-gecode-py) which uses the minizinc+gecode embedded in scikit-decide, and
```
poetry install --extras all
poetry run python postbuild_dev.py
poetry run python gecode.py
```

g-poveda (commented):
Hi I wonder if the small postbuild could be included during the build process itself, why the location of the files can't be done at this time? 
otherwise everything works fine on Mac OS

nhuet:
I tried that at first but did not find how to do that with poetry. So this is the best solution i found. (Actually other people wanted to do that during building process and poetry people answered that it was not possible but i do not remember the post now)

g-poveda (approved):
Working on all platform, no pb

Issue #192, opened 2022-05-12
[ci/cd] Build doc does not work anymore


nhuet:
## 🐛 Bug

The "build doc" step from "build scikit-decide" workflow is not working anymore.

## To Reproduce

1. Wait for scheduled built
1. Look for corresponding run

See for instance https://github.com/airbus/scikit-decide/actions/runs/2310818964

## Expected behavior

Doc built as before.

## Leads

This comes probably from a new version in dependencies.

nhuet:
Actually this is due to the installation of preleases **even for dependencies** (here protobuf 4.12-rc) via
```
pip install --pre --find-links ./wheels/ scikit-decide[all]
```
This was necessary so that --find-links option ./wheels/ accept to install the "prerelease" version of scikit-decide downloaded from build step. But it allows also downloading prerelease versions of dependencies.

So a workaround would be to specify explicitely the path to the proper wheel to install. (Using wildcards and specifying only plateform and python version). As in:
```
pip install wheels/scikit_decide*cp37*manylinux*.whl
```

PR #193, opened 2022-05-30
Avoid installing prereleased dependencies in github action

nhuet:
The way the scikit-decide library was installed for tests was designed
so that pip find scikit-decide (in prelease mode) in the "wheels" directory.
However a side effect was that it also allowed pip to install preleased
dependencies, which can lead to some bugs, see issue #192.

To avoid this we rather specify explicitely the path to the wheel (but
unsing wildcards to get easily the exact name) and get rid of the
"--pre" option.

We do that in a separate step for better readibility.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -378,6 +378,9 @@ jobs:
         compiler: [gnu]
       fail-fast: true
     runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


nhuet:
Actually the doc build is still broken because the dependency that was causing trouble (protobuf which is itself a dependency of ray) is now pass from prerelease to release status. See https://github.com/protocolbuffers/protobuf/issues/10051 and https://github.com/ray-project/ray/issues/25282 for more details.

We could pin the version of protobuf < 4 (in another PR) as suggested in the link , but we could also just wait for https://github.com/ray-project/ray/issues/25282 to be resolved. The ray team seems indeed rather reactive and have very frequent releases (~ each month start).

nhuet:
> Please pin protobuf until ray is fixed.

Done in #196.

PR #203, opened 2022-08-29
Update pynng dependency for windows

nhuet:
There is no wheel for pynng 0.6.2 for python 3.9 on windows.
So when installing, the wheel has to be generated and it makes
crash the github workflow since 24/06/2022.

See https://github.com/airbus/scikit-decide/runs/7035048822?check_suite_focus=true


galleon (approved):

galleon:
@fteicht shall we move at some point to GRPC ? What is the impact on C++ part ?

PR #204, opened 2022-09-18
set_window_title removed in last Matplotlib version

galleon:
The set_window_title function was deprecated in Matplotlib 3.4 and has be removed in 3.6


dbarbier:
There is another occurrence in examples/maze_multiagent_mdp_multisolve.py; in my opinion it would be simpler to just remove this line, examples are there to explain how to use scikit-decide, not matplotlib.

neo-alex:
PR discarded in favour of PR #206

Issue #175, opened 2022-01-12
GymDiscreteActionDomain does not discretize Box space correctly


galleon:
## 🐛 Bug

GymDiscreteActionDomain does not discretize Box space correctly

## To Reproduce

Steps to reproduce the behaviour:

1. I have a gym env which define a Box(low=-1, high=1, shape=(2,)) action space
2. When using IW which uses a custom wrapper inheriting from GymDiscreteActionDomain
3. The action space is not properly discretized as the `domain._applicable_actions.get_elements()` contains `[-1., 1.875]`which is not a valid action

## Expected behavior

I expected all generated elements to be in `[-1, 1]^2`

nhuet:
Should be closed as duplicated with #205 which has been solved.

fteicht:
Solved in PR #205

PR #205, opened 2022-09-18
Change way action space is discretized 

galleon:
Was PR #176 related to issue #175 which was closed 

This commit change the way a Box action is discretized.
It is inspired from the gym.spaces.Box.sample method. For unbounded spaces it builds on the quantile distribution of the Cauchy distribution (using a scale of 1).
An alternative could be derived from the logit function: log(1+p) - log(1-p) for example. But the current implementation has a larger spread.

### DIFF
--- a/skdecide/hub/domain/gym/gym.py
+++ b/skdecide/hub/domain/gym/gym.py
@@ -12,6 +12,8 @@ from collections import (  # TODO: replace with `from typing import NamedTuple`?
     namedtuple,
 )
 from copy import deepcopy
+from itertools import product
+from math import pi, tan
 from typing import Any, Callable, List, Optional
 
 import gym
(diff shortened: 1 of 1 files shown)


skdecide/hub/domain/gym/gym.py
```diff
@@ -700,30 +702,50 @@ def _discretize_action_space(
...
-                            for i in range(self._discretization_factor)
+            ticks = []
+
+            unbounded = ~action_space.bounded_below & ~action_space.bounded_above
+            upp_bounded = ~action_space.bounded_below & action_space.bounded_above
+            low_bounded = action_space.bounded_below & ~action_space.bounded_above
+            bounded = action_space.bounded_below & action_space.bounded_above
+
+            it = np.nditer(action_space.low, flags=["multi_index"])
+            for p in it:
```
fteicht, line 713: Variable `p` is not accessed, is it intended? If yes, please use another iteration scheme.

fteicht (changes requested):
Apart from variable `p` being not accessed in Line 713 of skdecide/hub/domain/gym/gym.py, it all sounds good to me!

skdecide/hub/domain/gym/gym.py
galleon: @fteicht modification & rebase done. Tests are all good.

fteicht (approved):

PR #207, opened 2022-09-20
Replace try..except by an method existence check to ease debugging

nhuet:
The call to init_model() was inside a try.. except block because some solver does not have such method.
We replace this block by a test wether the method exists so that an error inside the init_model() will be seen to better understand why a solve does not work as expected.


g-poveda (approved):
ok, better way of dealing with init_model in discreteopt solvers.

PR #208, opened 2022-09-23
Use independent discrete-optimization library

nhuet:
The discrete-optimization part of scikit-decide is now managed in an independent [repository](https://github.com/airbus/discrete-optimization).
The dependencies are updated to use the newly deployed library and the code is slightly updated to adapt to the changes in [discrete-optimization 0.2.0](https://pypi.org/project/discrete-optimization/0.2.0/).


galleon (commented):
LGTM

PR #211, opened 2022-09-26
Add a tasks_full_details attribute to State for scheduling notebook

nhuet:
After the code optimization made on scheduling algorithms in commit b76e15a86, the attribute `tasks_details` does not keep track of all tasks anymore. Thus:

- we introduce a new attribute `tasks_full_details` which holds also information for non-completed tasks and is actually a property reconstructed on-the-fly via `rebuild_all_tasks_dict`
- we add a dummy task for each task id in `rebuild_all_tasks_dict` output so that we can call a task detail even before starting any task as required in the scheduling notebook

### DIFF
--- a/notebooks/13_scheduling_tuto.ipynb
+++ b/notebooks/13_scheduling_tuto.ipynb
@@ -222,7 +222,7 @@
    "outputs": [],
    "source": [
     "print(\n",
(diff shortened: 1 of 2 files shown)


galleon (approved):
LGTM

PR #212, opened 2022-09-26
Build macos wheel on macos-latest github runner

nhuet:
Macos-10.15 github runner is deprecated and will be removed starting December (cf https://github.com/actions/runner-images/issues/5583). So we need to move the build on macos-latest (= macos-11 currently) and see if tests still pass on macos-10.15.

The other github runner names are also change for their "$os-latest" aliases when possible to follow more easily the update of github runners.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
         run: |
           python_version = ["3.7", "3.8", "3.9"]
           build = [ "macos-latest", "ubuntu-latest", "windows-latest" ]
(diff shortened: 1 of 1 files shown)


galleon (changes requested):
Ok for builds done on `latest` but I think tests should still using exact os names `macos-10.15`, `macos-11` and `macos-12`

nhuet:
I had to work a bit so that the wheel still works on macos 10.15 (it needs at least to have the proper name). I rebase with the solution i found and let the tests as before (not mentioning **-latest in tests oses)

nhuet:
Nice! The workflow passed this time! And with the requested changes.

galleon (approved):

PR #213, opened 2022-09-29
Remove embedded minizinc

nhuet:
Minizinc is actually a dependency of [discrete-optimization](https://github.com/airbus/discrete-optimization) and thus is managed by this project.
Thus we removed building process for libminzinc, chuffed, and gecode and configuring the minizinc binary path in `skdecide.hub`.

In discrete-optimization, it has been chosen that the user has to install manually minizinc binary and add it to PATH following the [minizinc documentation](https://www.minizinc.org/doc-latest/en/installation.html) (usually 2-4 lines of command lines in a terminal), so we add this step in installation instructions and in github actions workflow.


galleon (approved):
LGTM

PR #214, opened 2022-10-06
Correct install doc

nhuet:
There were an error on virtual environment activation on linux -> source skdecide-venv/bin/activate

And the proposed Python version was not installing correctly with conda.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -13,29 +13,29 @@ The use of a virtual environment for scikit-decide is recommended, and you will 
 This can be achieved either by using [conda](https://docs.conda.io/en/latest/) or by using [pyenv](https://github.com/pyenv/pyenv) (or [pyenv-win](https://github.com/pyenv-win/pyenv-win) on windows)
 and [venv](https://docs.python.org/fr/3/library/venv.html) module.
 
-The following examples show how to create a virtual environment with Python version 3.8.11 with the mentioned methods.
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -13,29 +13,29 @@ The use of a virtual environment for scikit-decide is recommended, and you will
...
 and [venv](https://docs.python.org/fr/3/library/venv.html) module.
 
-The following examples show how to create a virtual environment with Python version 3.8.11 with the mentioned methods.
+The following examples show how to create a virtual environment with Python version 3.8.13 with the mentioned methods.
 
 #### With conda (all platforms)
 
 ```shell
-conda create -n skdecide python=3.8.11
+conda create -n skdecide python=3.8.13
```
g-poveda, line 21: Is there a reason for that ?

g-poveda (commented):
just waiting an answer on the change of python default version ?

docs/install.md
nhuet: Yes, (as i said in the commit message ;p) the install was not working with conda (probably because of dependencies conflicts i do not now). But it worked with python 3.8.13 so i chose to change everywhere the python version in the example so that one can copy/paste and have it work.

g-poveda: my bad, didn't open the commit description. then lgtm

g-poveda (approved):
no pb

PR #215, opened 2022-10-11
Remove duplicate step uploading doc to github-pages

nhuet:
This deployment was made twice when pushing on master. It is better to do it only if all tests pass, so we keep the second instance of the step instead of the first one.


g-poveda:
lgtm

PR #216, opened 2022-10-12
Refresh workflows

galleon:
Refreshing all actions used in both build and release workflows in prevision of node12 deprecation

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,7 +91,15 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.8"
-      - uses: pre-commit/action@v2.0.0
+      - name: install pre-commit
+        run: python -m pip install pre-commit
+      - name: get cached pre-commit hooks
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
+      - name: pre-commit checks
+        run: pre-commit run --show-diff-on-failure --color=always
 
   build-windows:
     needs: [setup]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
       do_windows: ${{ steps.generate-matrix.outputs.do_windows}}
       build_doc: ${{ steps.generate-matrix.outputs.build_doc}}
     steps:
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v4
       - name: Generate Matrix
         id: generate-matrix
         shell: python3 {0}
@@ -87,8 +87,8 @@ jobs:
   lint-sources:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
         with:
           python-version: "3.8"
       - uses: pre-commit/action@v2.0.0
@@ -108,25 +108,25 @@ jobs:
 
     steps:
       - name: Checkout scikit-decide source code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           submodules: true
           fetch-depth: 0
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
 
       - name: Load cached venv
         id: cached-pip-wheels
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: ~/.cache
           key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
 
       - name: Restore Boost cache
-        uses: actions/cache@v2.1.4
+        uses: actions/cache@v3
         id: cache-boost
         with:
           path: ${{env.BOOST_DIR}}
@@ -143,7 +143,7 @@ jobs:
 
       - name: Restore build dependencies
         id: cache-build-dependencies
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: |
             skdecide/hub/bin
@@ -167,7 +167,7 @@ jobs:
         run: 7z x dist/*.whl -y
 
       - name: Upload as build artifacts
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: wheels
           path: dist/*.whl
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -423,7 +431,7 @@ jobs:
           echo "::set-output name=path::${{ env.minizinc_cache_path }}"  # expands variables
```
nhuet, line 431: You forgot to bump download-artifact version at lines 424, 504, 584, 684, 717
You have also to add path argument when not yet present to make it work:
```yaml
      - name: Download artifacts
        uses: actions/download-artifact@v3
        with:
          name: wheels
          path: wheels
```

.github/workflows/release.yml
```diff
@@ -677,7 +684,7 @@ jobs:
 
     steps:
       - name: Delete nightly release
-        uses: actions/github-script@v5
+        uses: actions/github-script@v6
         with:
           script: |
             const releases = await github.rest.repos.listReleases({
```
nhuet, line 690: You missed also bumping download-artifact to v3 in release.yml (in order to use actions with node 16 instead of node 12)

nhuet (commented):
You could also change JamesIves/github-pages-deploy-action version in build.yml (for v4 as in release.yml) so that:
- it is consistent with release.yml
- more importantly it use the last v4.x.y version, which is now using node 16

Personnally i would rather squash all commit "Advancing..." and add explanation in commit message that it is to get node 16 using actions because of node 12 end of support.

I would like also having a small explanation about why you drop pre-commit action and change the action creating the release. (I know why because I did the same for discrete-optimization but someone else might want to know the reason requiring these changes).

nhuet (approved):
LGTM

dbarbier:
There are still many warnings

nhuet:
> There are still many warnings

This is another matter: these warnings (at leas those I have seem) are not about node 12, but rather abour "`set-output` command is deprecated". And are actually coming from github actions (maintained by github) for which the version used is already the more recent. e.g. download-artifact@v3 leads to these warnings.

My guess is that github will soon update these actions and that we will have to bump them again at that time.

dbarbier:
> > There are still many warnings
> 
> This is another matter: these warnings (at least those I have seem) are not about node 12, but rather about "`set-output` command is deprecated". And are actually coming from github actions (maintained by github) for which the version used is already the more recent. For instance, download-artifact@v3 leads to these warnings.
> 
> My guess is that github will soon update these actions and that we will have to bump them again at that time.

No, these warnings come from our workflows, grep "set-output".

.github/workflows/build.yml
```diff
@@ -77,21 +84,36 @@ jobs:
                   build_doc = "false"
           oses = ["macos", "ubuntu", "windows"]
           build_dict = {os : [k for k in build if k.startswith(os)] for os in oses}
-          print(f"::set-output name=build::{build_dict}")
-          print(f"::set-output name=test::{dict({os : [k for k in test if k.startswith(os)] for os in oses})}")
-          print(f"::set-output name=build_doc::{build_doc}")
-          for os in oses:
-              print(f"::set-output name=do_{os}::{'true' if len(build_dict[os]) > 0 else 'false'}")
-          print(f"::set-output name=python_version::{python_version}")
+          with open(getenv("GITHUB_OUTPUT", default="/tmp/t"), "a") as f:
```
dbarbier, line 87: I do not get it, why don't you use GITHUB_ENV?

.github/workflows/build.yml
```diff
@@ -77,21 +84,36 @@ jobs:
...
+              f.write(f"build={build_dict}")
```
dbarbier, line 88: Add newlines everywhere:
```suggestion
              f.write(f"build={build_dict}\n")
```

.github/workflows/build.yml
dbarbier: Also since there is no reason to use default values, os.environ could be used instead of getenv.

.github/workflows/build.yml
galleon: Merci

.github/workflows/build.yml
galleon: The environment is different for each job so we have to go through the same mechanism and use GITHUB_OUTPUT

galleon:
The remaining warnings come from `actions/download-artifact@v3`. There is a PR to update to the latest `actions/core` that will solve this - not is our hands.

.github/workflows/build.yml
```diff
@@ -77,25 +82,35 @@ jobs:
...
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
+      - name: pre-commit checks
+        run: pre-commit run --show-diff-on-failure --color=always
 
   build-windows:
     needs: [setup]
-    if: needs.setup.outputs.do_windows == 'true'
+    if: needs.setup.outputs.do_windows
```
dbarbier, line 113: Are you sure that this change does not break anything?  AFAICT you did not test it.

.github/workflows/build.yml
```diff
@@ -413,32 +428,38 @@ jobs:
...
+        uses: actions/download-artifact@v3
         with:
           name: wheels
+          path: wheels
 
       - name: get MininZinc path to cache
         id: get-mzn-cache-path
         run: |
-          echo "::set-output name=path::${{ env.minizinc_cache_path }}"  # expands variables
+          echo "path=${{ env.minizinc_cache_path }}" >> $GITHUB_OUTPUT
```
dbarbier, line 439: Please do not remove comment, it is useful

dbarbier (commented):
IMHO you should check release.yml in your fork

.github/workflows/build.yml
```diff
@@ -77,25 +82,35 @@ jobs:
...
-          print(f"::set-output name=test::{dict({os : [k for k in test if k.startswith(os)] for os in oses})}")
-          print(f"::set-output name=build_doc::{build_doc}")
-          for os in oses:
-              print(f"::set-output name=do_{os}::{'true' if len(build_dict[os]) > 0 else 'false'}")
-          print(f"::set-output name=python_version::{python_version}")
+
+          with open(environ["GITHUB_OUTPUT"], "a") as f:
+              f.write(f"build={dumps(build_dict)}\n")
+              f.write(f"test={dumps(dict({os : [k for k in test if k.startswith(os)] for os in oses}))}\n")
+              f.write(f"build_doc={dumps(build_doc)}\n")
```
dbarbier, line 89: You are modifying the contents of these variables without giving any reason; to get the same values, dumps() should be called only for "build" and "test".

dbarbier (commented):
Looks good, but you must explain in the last commit message why you want to merge these changes.  You claim that set-output is unsafe, this is not clear to me whether this is true.

.github/workflows/build.yml
galleon: Moved back to the previous statement.

.github/workflows/build.yml
galleon: Pushed back command event though it is not expanding anything, just setting an env variable.

galleon:
> Looks good, but you must explain in the last commit message why you want to merge these changes. You claim that set output is unsafe, this is not clear to me whether this is true.
I have used GH comment

g-poveda (approved):
trust in previous reviews.

PR #217, opened 2022-10-13
Updating binder for flight planning notebook

galleon:
Adding `libeccodes-tools` as linux package
And the following python packages:
```
- cartopy>=0.21
- cdsapi
- openap
- pygeodesy
- xarray
```

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,16 +3,23 @@ channels:
   - pytorch
   - defaults
 dependencies:
+  - cartopy>=0.21
   - python=3.8
   - pip
   - pytorch::cpuonly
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -14,7 +14,7 @@ dependencies:
     - matplotlib
     - ipywidgets
     - ipympl
-    - scikit-decide[all]==0.9.5
+    - scikit-decide[all]==0.9.4
```
dbarbier, line 17: You must explain in commit message why you make this change

dbarbier (approved):

PR #219, opened 2022-10-21
Remove postbuild script and fix doc for contributing

nhuet:
This script was used to finish properly the configuration of minizinc when it was built by scikit-decide. 

Now that we are using an external installation, this is not necessary anymore.


g-poveda:
LGTM

g-poveda (approved):

Issue #220, opened 2022-10-21
Build failing when using commit message [ci: macos-12] [ci: python-3.8] [ci: python-3.9] [ci: windows]


galleon:
## 🐛 Bug

The build is failing in build doc because not linux wheels are generated (they are needed for the doc generation step)

## To Reproduce

1. Add a file and commit with a message including this [ci: macos-12] [ci: python-3.8] [ci: python-3.9] [ci: windows]

## Expected behavior

I expect the build to generate the doc on a platform that is requested

## Environment

All

galleon:
I have a fix but those options are rarely used. I believe they were useful when the solvers were build but might be less useful today. Do we keep them?

PR #221, opened 2022-10-21
Generate doc on a platform for which the wheels have been build

galleon:
Make sure the documentation is always generated on a platform for which the wheel have been built and tested. Fix for #220


g-poveda:
to be reopened later if needed

PR #223, opened 2022-11-22
Update minizinc install command line for ubuntu in github actions

nhuet:
Works even when FUSE is not available.
Avoid installing minizinc for building doc by setting DO_SKIP_MZN_CHECK environment variable for discrete-optimization.


g-poveda (approved):
Ok lgtm, coherent with what has been done in discreteoptim

PR #224, opened 2022-11-22
Add tips in doc to install minzinc easily in a terminal (works on cloud)

nhuet:


g-poveda (approved):
Ok coherent with discropt documentation too thanks

PR #226, opened 2022-11-22
Install last nightly build in binder env for master version

nhuet:
Here we set manually the url of the nightly build. The github action will be modified so that
- a new nightly build will update this url
- the environment for a release will empty this url

Note that we do not touch the environemnt.yml so that previous release of scikit-decide should already been installed with all its dependencies in docker cache.
It should speed-up the new image build as only updated dependencies should be downloaded.


nhuet:
The new generated environment can be tested here: [Binder](https://mybinder.org/v2/gh/nhuet/scikit-decide/nht/binder_nightly)

If you type 
```python
import skdecide
import discrete_optimization
print(skdecide.__version__)
print(discrete_optimization.__version__)
```
In a cell of a new notebook, you will see that the master version of scikit-decide has been installed (with the separate discrete-optimization library which was not the case for scikite-decide 0.9.4)

PR #230, opened 2022-11-23
Fix missing variable python_version for skdecide install in release.yml

nhuet:
The installation of scikit-decide did not work during tests for release workflow. 
This fixes a bug introduced by a947139.


g-poveda (approved):
Ok

PR #231, opened 2022-11-24
Update dependencies to use last discrete-optimization version

nhuet:
Updated pyproject.toml and then applied `poetry update` to update accordingly poetry.lock.


g-poveda (approved):
Approved, but it may be questionable to put the poetry.lock in git ? to be investigated @fteicht

Issue #236, opened 2023-01-14
Implement a void rollout policy in MCTS


fteicht:
## 🚀 Feature

Having the choice to voice the rollout policy in MCTS, i.e. not calling the rollout policy, but relying instead on default leaf node values or - better - on heuristic estimates of the leaf nodes.
Heuristic estimates of leaf nodes is already an existing option in scikit-decide's MCTS implementation, but not the possibility to voice the rollout policy.

## Motivation

Rollout policies can be highly CPU demanding when sampling is costly.
In this case, it makes more sense to replace the rollout policy with a heuristic estimate of the leaf nodes in the MCTS tree, and to not call the rollout policy.

## Pitch

When creating the MCTS solver, having the possibility to use `rollout_policy=MCTS.Options.RolloutPolicy.Void` as a new option.
Should raise a warning if the `heuristic` option is not set as well, in which case the default node values (i.e. 0) is used.

## Alternatives

There is no alternative to fake the void rollout policy properly.

PR #237, opened 2023-01-15
236 implement a void rollout policy in mcts

fteicht:
Implement issue #236


g-poveda:
The modification seems good, it would be nice to have some associated unit test

PR #241, opened 2023-02-17
Update isort version to make it work with newer version of pre-commit

nhuet:
The last version of pre-commit is failing to install the isort hook for the older version. It is now working properly.

You may actually not have noticed this issue due to pre-commit cache. It appears after a  ̀pre-commit clean` or in the actions if the cache for the hooks is not hit.


g-poveda (approved):
Ok, this problem has been identified in other repo with the same fix afaik so LGTM

PR #242, opened 2023-02-17
Upper-bound poetry-core version when building the library

nhuet:
This is a quick fix to make the current process work. A better solution should be found afterwards to avoid limiting the dependencies with upper bounds.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -123,5 +123,5 @@ profile = "black"
 all = true
 
 [build-system]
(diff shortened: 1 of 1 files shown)


g-poveda (approved):
Temporary solution seems ok, can you put an issue @nhuet to keep track of it. Maybe there is a very fast way to raise an issue from your comment in the PR

g-poveda:
Ok issue raised #243

Issue #243, opened 2023-02-17
Poetry >=1.5 ignore some options


g-poveda:
#242 PR is a quick fix that permits that the c++ library is still built. A better solution should be found afterwards to avoid limiting the dependencies with upper bounds.
the part that build c++ lib is this one and seems to be ignored with latest version of poetry
```
[tool.poetry.build]
script = "builder.py"
```

g-poveda:
solved by #245

PR #245, opened 2023-02-17
Fix library build with last version of poetry-core

nhuet:
By adding an argument to "tool.poetry.build" section, we go back to previous behaviour and make the build process work with poetry-core 1.15


g-poveda:
Ok this seems to fix #243 thanks

g-poveda (approved):
Ok Closes #243

PR #249, opened 2023-02-23
Build wheels for python 3.10

nhuet:


g-poveda (approved):
Looks good, so there is no problem with py3.10 it seems..!

PR #250, opened 2023-03-07
Fix release workflow

nhuet:
We fix
- the build library step by upgrading poetry-dynamic-versioning so that the proper version number is used for the generated wheel
- the build doc step by specifying properly the wheel to install


nhuet:
The resulting release workflow is available [here](https://github.com/nhuet/scikit-decide/actions/runs/4355317016) (corresponding to a fake release v0.1.3 on my fork).
Note in particular the names of the wheels generated.

[workflow still in progress]

g-poveda (approved):
lgtm, solve a problem in workflow.

PR #253, opened 2023-04-20
Preinstall gym==0.21.0 with option `--no-use-pep517`

nhuet:
New versions of pip/setuptools enforce pep 517 by default, which is not respected by gym 0.21.0 install. 
So we preinstall with option `--no-use-pep517` so that the installation of `scikit-decide[all]` works.

Note that it could be also nice to upgrade to gymnasium 0.26 and later but this is not yet possible because of a conflict between recent versions of [`ray[rllib]`](https://github.com/ray-project/ray/issues/34396) and [`stable-baselines3`](https://github.com/DLR-RM/stable-baselines3/issues/1452) which are each fixing a different version of gymnasium.

We also add this tip in the documentation to install scikit-decide (normal process and developper mode).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -271,7 +271,7 @@ jobs:
           # cross-compile for macosx-10.15
           export MACOSX_DEPLOYMENT_TARGET=10.15
           python -m build --sdist --wheel "--config-setting=--plat-name=macosx_10_15_x86_64"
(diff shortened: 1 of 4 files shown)


g-poveda (approved):
It looks good. I don't like this kind of fix but we are not responsible right ?!, fine for me

fteicht (approved):
Seems good to me, thanks!

PR #258, opened 2023-06-14
Flight Planning Domain

YoanwM:
Creation of a Flight Planning Domain for scikit-decide. 
It contains wind interpolator tools, its own README & notebook, some utils and the domain.


fteicht (changes requested):
Thanks for the great pull request!
Before accepting it, we would need:

1. that you merge all the commits in only 1 commit
2. that you include an assert in your unit test to check that the domain is compatible with the solver and that no exception is thrown.

Thanks!

PR #260, opened 2023-06-29
Update gym 0.21.0 install by bounding setuptools version

nhuet:
Gym 0.21.0 requirement list is broken and the install process cannot work anymore with latest version of setuptools.

We update tips in documentation and install process in github actions.


fteicht (approved):
Seems correct for me, thanks @nhuet !

PR #261, opened 2023-06-29
Change dependencies between github actions jobs

nhuet:
Sometimes, if not waiting for all builds to finish, the artifact download in test steps may fail if it happens approximately at the same time another build finishes. (The artifact is then supposed to contain wheels not yet fully uploaded)


g-poveda (approved):
lgtm

PR #267, opened 2023-09-09
Update to latest releases of cpp sdk submodules

fteicht:
All the cpp sdk submodules have been updated to their latest official release.
It notably includes a bug fix in pybind11 that enables to compile on Windows with Microsoft Visual Studio Build Tools 2022.


neo-alex (approved):
I am no CPP specialist, but I see no problem to merge.

PR #270, opened 2023-09-14
Synchronize package dependencies between build.yml and release.yml

fteicht:
Package dependencies were recently changed in build.yml but not in release.yml, which makes the release process fail.


neo-alex (approved):
Seems all good for merging!

PR #273, opened 2023-10-09
Add --all-files option to pre-commit command to avoid skipping the test

nhuet:
No tests were actually done by the linter action because `pre-commit run` look only files in the index by default.
When checkouting a whole PR, no files are in the index so no files are checked.

Previously, the job returned something like

    Trim Trailing Whitespace.............................(no files to check)Skipped
    Fix End of Files.....................................(no files to check)Skipped
    Check Yaml...........................................(no files to check)Skipped
    Check for added large files..........................(no files to check)Skipped
    Pretty format JSON...................................(no files to check)Skipped
    isort (python).......................................(no files to check)Skipped
    black................................................(no files to check)Skipped
    nbqa-isort...........................................(no files to check)Skipped
    nbqa-black...........................................(no files to check)Skipped
    nbqa-pycln...........................................(no files to check)Skipped
    nbstripout...........................................(no files to check)Skipped

Now the checks are done (and should read as "Passed") 
We make a `pre-commit run --all-files` in this commit to have the job pass.

We also remove the pretty-format-json hook which was tampering with the notebooks (oddly only remotely on the github runner, not locally on my laptop), and was not very useful for the repository anyway.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,7 +105,7 @@ jobs:
           path: ~/.cache/pre-commit
           key: pre-commit|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
       - name: pre-commit checks
-        run: pre-commit run --show-diff-on-failure --color=always
+        run: pre-commit run --show-diff-on-failure --color=always --all-files
 
   build-windows:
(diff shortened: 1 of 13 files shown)


g-poveda (approved):
lgtm

PR #275, opened 2023-10-09
Correct CGP behaviour regarding verbosity and Restorable API

### LINKED ISSUE #177: CGP violates interface & options
## 🐛 Bug

When using `verbose=False` CGP still outputs some info for each iteration. The default number of iterations being 1_000_000 this quickly lead to unpleasant output.

`CGPWrapper`  subclass `Restorable` but does not implement neither `_load` nor `_save`

## To Reproduce

1. You can use the AI Gym notebooks for which iterations is only 25.

## Expected behavior

No output when `verbose=False`

nhuet:
This PR fixes #177.

- We remove all prints when verbose=False
- We remove inheritance from Restorable since CGPWrapper did not implement `_save()` nor `_load()`.

To see how verbosity is fixed you can compare how the following snippets behave before and after the PR:

```python
import gym
from skdecide.hub.domain.gym import (
    GymDomain,

)
from skdecide.hub.solver.cgp import CGP


ENV_NAME = "MountainCarContinuous-v0"
domain_factory = lambda: GymDomain(gym.make(ENV_NAME))

print("*** With verbosity")
solver = CGP("TEMP_CGP", n_it=2, verbose=True)
GymDomain.solve_with(solver, domain_factory)

print("***Without verbosity")
solver = CGP("TEMP_CGP", n_it=2, verbose=False)
GymDomain.solve_with(solver, domain_factory)

```

### DIFF
--- a/skdecide/hub/solver/cgp/cgp.py
+++ b/skdecide/hub/solver/cgp/cgp.py
@@ -303,6 +303,7 @@ class CGPWrapper(Solver, DeterministicPolicies, Restorable):
             evaluator,
             self._folder_name,
             self._n_cpus,
+            verbose=self._verbose
         )
         es.run(self._n_it)
 
--- a/skdecide/hub/solver/cgp/cgp.py
+++ b/skdecide/hub/solver/cgp/cgp.py
@@ -201,7 +201,7 @@ def denorm(vals, types):
         return out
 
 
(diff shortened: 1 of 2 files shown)


neo-alex (approved):
OK for me, thanks

PR #277, opened 2023-11-10
github actions: do not wait for other os to build before launching an os test suite

nhuet:


g-poveda (commented):
was there a reason we were doing the previous behaviour ? maybe there were some random problem sometimes ?

nhuet:
Not that I am aware of. It was designed by @galleon, maybe he can enlight us?

fteicht (approved):

PR #284, opened 2023-11-24
Fix ray_rllib solver after the upgrade to ray 2.7.0

nhuet:
The `RayRLlib` solver class wraps `ray.rllib` algorithms.
Since we upgrade to gymnasium, the dependency in ray also upgrade from
1.10 to >=2.7.0 and the rllib api changed a bit.

Main changes/fixes:
- `ray.rllib.agents` becomes `ray.rllib.algorithms`, `Trainer` -> `Algorithm`
- `policy_mapping_fn`: maps now agent_id + episode + worker -> policy_id
- `algo.compute_action` -> `compute_single_action()`
- `MultiAgentEnv` follows now gym0.26/gymnasium convention => possibility
  to use `MultiAgentEnvCompatibility` rllib wrapper. We choose here,
  consistently with the work done in `skdecide.hub.domain.gym` to
  keep the old class `AsRLlibMultiAgentEnv` with the gym 0.21 api and wraps it.
  More precisely:
    - the old env is renamed as `AsLegacyRLlibMultiAgentEnv` and derives from `AsLegacyGymV21Env`
      in order to share some  methods and define some missing ones (as `render()`, `seed()`, `close()`)
    - the new `AsRLlibMultiAgentEnv` derives from `MultiAgentEnvCompatibility` which itself
      derives from `MultiAgentEnv`
- `MultiAgentEnv` needs a `get_agent_ids()` (or at least a `_agent_ids`
  attribute)
- `MultiAgentEnv` needs to define `observation_space` and `action_space`
  attributes. Preferably as `gym.spaces.Dict` in order to allow implicit
  implementation of new methods `action_space_sample()` and `observation_space_sample()`
- `done` output in `step()` must be a dictionary with a key by agent + a
  `"__all__ "` key
- `algo.load()` takes directly the directory path (which is overwritten at
  each store)

Other fixes done here:
- `__init__()` of `RayRLlib` was using a mutable object as default value for
  policy_config which can lead to unforeseen bugs
- main script moved as test in tests/ together with a test dedicated to
  `AsRLlibMultiAgentEnv`
- fixes `"__all__"` value of done dictionary as `MultiAgent` domains are
  supposed to returned a dictionary agent_id->boolean instead of a
  single boolean
- update the examples/rllib_solver.py
- test also single agent domain to check the autocast works properly
  with rllib


tests/solvers/python/test_ray_rllib.py
```diff
@@ -0,0 +1,120 @@
...
+        assert isinstance(subspace, gym.spaces.Space)
+
+    # checks from ray.rllib
+    ray.rllib.utils.check_env(env)
+
+
+def test_as_rllib_env_with_autocast_from_singleagent_to_multiagents():
+    ENV_NAME = "CartPole-v1"
+
+    domainupcasted = GymDomain(gym.make(ENV_NAME))
```
neo-alex, line 36: As in the previous PR, I suggest renaming it `upcast_domain`

tests/solvers/python/test_ray_rllib.py
```diff
@@ -0,0 +1,120 @@
...
+def test_ray_rllib_solver():
+    # define domain
+    domain_factory = lambda: RockPaperScissors()
+    domain = domain_factory()
+
+    # check compatibility
+    assert RayRLlib.check_domain(domain)
+
+    # solver factory
+    config_factory = lambda: PPO.get_default_config().resources(
```
neo-alex, line 65: A config_factory seems overkill here (why not just `config`?)

neo-alex (approved):
Thanks, please consider the few suggested minor changes and then LGTM.

tests/solvers/python/test_ray_rllib.py
nhuet: I added a comment in the code to explain it.

neo-alex: Oh OK, and then I suppose that the config being frozen is an issue (even if we don't modify it?)...

PR #285, opened 2023-11-27
Update python and scipy versions

nhuet:
- Force a scipy version (>=1.7.3) with wheels available for 3.10 (and 3.9, 3.8, and 3.7)
- Upper bound python with <3.11 so that poetry is ok to install scipy>=1.7.3 (which wants also python>=3.7,<3.11)

NB: gymnasium 0.28.1 also is tested with scipy 1.7.3

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -86,74 +86,88 @@ files = [
 
 [[package]]
 name = "certifi"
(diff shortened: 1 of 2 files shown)


nhuet:
Outdated by #287

PR #287, opened 2023-11-27
Update python allowed: [3.8 ; 3.11]

nhuet:
We start now from 3.8 and end with 3.11.
We build a wheel for 3.11, except on windows. The reason for this exception is that there is dependency conflicts when installing scikit-decide[all] on windows python3.11 (seemingly due to pyarrow) and thus we could not test the generated wheel.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       python_version: ${{ steps.generate-matrix.outputs.python_version }}
+      python_version_per_os: ${{ steps.generate-matrix.outputs.python_version_per_os }}
       build: ${{ steps.generate-matrix.outputs.build}}
       test: ${{ steps.generate-matrix.outputs.test}}
       do_macos: ${{ steps.generate-matrix.outputs.do_macos}}
@@ -81,6 +82,13 @@ jobs:
                   build_doc = "false"
           oses = ["macos", "ubuntu", "windows"]
           build_dict = {os : [k for k in build if k.startswith(os)] for os in oses}
+          python_version_per_os = {os: python_version for os in oses}
+          # remove python 3.11 for windows: dependency conflict from pyarrow prevent testing the wheel windows python 3.11
+          python_version_per_os["windows"] = [v for v in python_version if v != "3.11"]
+          # update build_dict by removing os without any python version
+          for os in build_dict:
+              if len(python_version_per_os[os]) == 0:
+                    build_dict[os] = []
 
           with open(environ["GITHUB_OUTPUT"], "a") as f:
               f.write(f"build={build_dict}\n")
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -81,6 +82,13 @@ jobs:
                   build_doc = "false"
           oses = ["macos", "ubuntu", "windows"]
           build_dict = {os : [k for k in build if k.startswith(os)] for os in oses}
+          python_version_per_os = {os: python_version for os in oses}
+          # remove python 3.11 for windows
```
neo-alex, line 86: What is the issue with Python 3.11 on Windows (could it be added to this comment to not forget in the future and fix when possible)?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "scikit-decide"
-version = "0.9.8.dev19+4ad63a353"
+version = "0.0.0"
```
neo-alex, line 3: Why version "0.0.0"?? Do you mean "0.10.0"?

pyproject.toml
```diff
@@ -50,9 +50,10 @@ format-jinja = """
 files = ["skdecide/__init__.py"]
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = ">=3.8,<3.12"
```
neo-alex, line 53: I am wondering is ">=3.8,<3.12" is better than "^3.8" (to avoid changing this line regularly)? I see pros & cons...

neo-alex (commented):
Thank you for this PR, please check the few comments added.
Also, I think it would be great updating the parts of the documentation mentioning Python 3.7 e.g. https://airbus.github.io/scikit-decide/install.html#python-3-7-environment

pyproject.toml
nhuet: Actually this line is updated dynamically at build (via poetry-dynamic-versioning) and should not be changed by hand. The fact that it was modified here to a specific version was a previously introduced error. But I agree it should not specially be done in this PR, it is a bit out of scope.

.github/workflows/build.yml
nhuet: There is a conflict version on dependencies (I think coming from pyarrow) with python >=3.11. Thus we cannot install scikit-decide[all] and test the windows python3.11 wheel. So I chose not to publish this wheel for now. We could also decide to publish it untested but I am not sure this is really a good idea.

nhuet: I added the reason on the PR description. Is it sufficient for you or do you prefer i put it also in build.yml?

pyproject.toml
nhuet: I removed it

nhuet:
All comments taken into account!

.github/workflows/build.yml
neo-alex: Thanks for the explanations, it is fine for me as you did.

pyproject.toml
neo-alex: OK I see!

PR #292, opened 2023-12-04
Ensure pybind11 finds the right Python interpreter

fteicht:
A change in the recent CMake policies prevents Pybind11 to use the Python interpreter given in scikit-decide's configure args via `-DPYTHON_EXECUTABLE=/path/to/python-interpreter`.

See [pybind11's issue #4785](https://github.com/pybind/pybind11/issues/4785) for more information.

This PR enforces Pybind11 to conform to cmake's CMP0148 policy and to use the python interpreter given to scikit-decide's configuration command line. This is critical since otherwise scikit-decide's build script might not create the wheel for the intended python version.


fteicht:
It happens to be a wrong configuration on my side. The default behaviour of pybind11 works like a charm whereas the line which this PR adds only make things more complex and it requires more work to actually work.

PR #293, opened 2023-12-05
Skip test_up_bridge_solver_numeric on MacOS

nhuet:
This is temporary because of a crash during UP `solve()`.


nhuet:
Not needed thanks to #295

PR #294, opened 2023-12-06
Reactivate multi-threading on MacOS wheels and update code to C++20 standards

fteicht:
This PR:
- Reactivate multi-threading when building MacOS wheels, which was deactivated because of cmake not able anymore to find OpenMP on MacOs without hints on where the libomp library is ;
- Re-separate pytests in separate processes, one per pytest folder, in order to avoid launching all the pytests in a same process and having OpenMP library version conflicts with some distributions ;
- Move the C++ codebase to c++20 standards which most dependent packages under the `cpp/sdk` folder are following.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -477,9 +477,17 @@ jobs:
           ${{ env.minizinc_config_cmdline }}
           # test minizinc
           python -c "import minizinc; print(minizinc.default_driver.minizinc_version); minizinc.Solver.lookup('gecode')"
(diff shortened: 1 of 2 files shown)


nhuet:
@fteicht Can you develop why you need to split again the tests ? (by the way you forgot the flight_planning tests: that is the risk, when specifying manually the subdirectories)

nhuet:
If I understand well, what is important is to call the cpp/ subfolders in an other process than the other directories (python/) ? Is it correct (and sufficient) ?

fteicht:
Yes you got it right @nhuet

g-poveda (approved):
lgtm !

PR #298, opened 2023-12-12
Fix release.yml

nhuet:
We do not want to publish a windows 3.11 wheel because of a dependencies conflict preventing to test it, as explained in commit 4767f4e.


g-poveda:
Hi @nhuet can you remind me which hard dependency need PyArrow library ? i don't see pyarrow in the toml so i guess it's an indirect dependency

nhuet:
By taking a look to poetry.lock, it seems to be ray[rrlib]

g-poveda (approved):

PR #300, opened 2023-12-12
Remove simplejson dependency

nhuet:
This was used as an optional way to dump episode in json format in rollout, but never really tested or even used.

We remove this feature from rollout from now and thus remove the dependency.


g-poveda (approved):
Look good to me

PR #303, opened 2024-01-02
Build wheels for macosx-arm64

nhuet:
We adapt the workflow used by scikit-learn (see this [workflow](https://github.com/scikit-learn/scikit-learn/blob/main/build_tools/wheels/build_wheels.sh) and this [script](https://github.com/scikit-learn/scikit-learn/blob/main/.github/workflows/wheels.yml)).

- We cross-compile from macosx-x86_64 (the only macosx github runner available).
- We use cibuildwheel to create the wheel, with the appropriate flags.
- We use a suitable version of openmp, taken from conda-forge and adapted to scipy.
    
Warnings:    
- The version of openmp used here is 11.1.0 (until now, brew was installing 17.0.6).
- The arm64 wheel is not tested later in the workflow as no github runner exists with macosx-amr64 platform.


fteicht (approved):
Thanks a lot @nhuet !

nhuet:
I solved the merge conflict @fteicht

PR #309, opened 2024-01-24
Improve unified planning support

fteicht:
This pull request improves the support of [unified planning](https://github.com/aiplan4eu/unified-planning) in scikit-decide, notably enabling to solve UP domains with the RLlib solver. It relies on:
- the recent handling of state-based applicable actions, which are produced by unified planning domains, in the RLlib solver ;
- the automatic translation to/from unified planning states and actions from/to numpy arrays that can be handled by RLlib.


g-poveda (commented):
I don't understand those commit, not related to up isn't it ?
is it some results of pre-commit or refactor you made ?

tests/domains/python/test_up_bridge_domain.py
```diff
@@ -9,15 +9,14 @@
 
 
 @pytest.mark.skipif(sys.version_info < (3, 10), reason="requires python3.10 or higher")
-def test_up_bridge_domain():
+def test_up_bridge_domain_random():
     noexcept = True
 
     try:
```
g-poveda, line 15: it can be the opportunity to remove the try except here ?

skdecide/hub/space/gym/gym.py
```diff
@@ -420,7 +420,7 @@ def get_elements(self) -> Iterable[T]:
         return self._elements
 
     def sample(self) -> T:
-        return self._elements[super().sample()]
+        return self._to_elements[super().sample()]
```
g-poveda, line 423: is it a bug you found during this PR implementation ?

fteicht:
> I don't understand those commit, not related to up isn't it ? is it some results of pre-commit or refactor you made ?

There are 2 kinds of commits here:
- Python imports reordering in order to comply with linting rules for Python
- Adding missing solvers and domains to `pyproject.toml` which I spotted when adding the UP bridge ones

skdecide/hub/space/gym/gym.py
fteicht: Correct, I found this bug when calling the `sample` method of `SetSpace` in the UP bridge test units.

g-poveda (changes requested):
See my previous comment :)

g-poveda (approved):

PR #318, opened 2024-04-02
Poll-Schumann model + plot + fixes

nestorcarmona:
- Added Poll-Schumann aircraft performance model as a Python library which can be pip installed (see here [poll_schumann](https://github.com/nestorcarmona/poll_schumann))
- Added poll_schumann to pyproject.toml (updated .lock file as well)
- Modified plot function for flight planning to include Zp and GW profiles + trajectory
- Modified notebook flight_planning_tuto.ipynb


nhuet:
Hi. Once again be careful to remove "__skdecide_hub_cpp.cpython-38-darwin.so" from the commits!

nestorcarmona:
Last commit fixed *.so file :) Added Poll-Schumann back as part of the repository + fixed tests

fteicht (changes requested):
Thanks a lot @nestorcarmona , it seems to me that it is a solid and very useful PR!

Three small comments on my side:
- In the flight planning tutorial notebook, you could change the section name "Solving and rendering out the flight planning" to "Solving and rendering out the flight planning domain"
- In the flight planning tutorial notebook, the last cell is empty and thus useless (it should be removed)
- In the .gitignore file, I am in favour of removing the commented line (on C extensions) and the ".so" line, because they are specific to your own development environment; if you want, you can keep the modified .gitignore file as such in your local branch without pushing it to the main repository, so that it will filter out the files you want without impacting the .gitignore file shared by everyone.

Also, there is a conflict with the poetry.lock file that should be resolved on your side.
Thanks again, you're nearly there!

nestorcarmona:
Thanks @fteicht ! I just pushed the requested corrections.

fteicht (approved):

PR #320, opened 2024-04-02
Add a solver characteristic for solvers implementing `solve_from()`

nhuet:
The new characteristic, the generic (default) version and its optional specific declination are:

- generic characteristic: FromInitState
- specific characteristic: FromAnyState:
   - the solver have a solve_from() method
   - solve() use it by default to solve from initial state (if domain is initializable else raise a ValueError if not implemented)
   - solve() (and _solve) gets also an optional argument from_memory to solve from another state (and call solve_from accordingly)

Consequences:
- add from_memory arg in Domain.solve_with()
- Solver inherits from FromInitState (to have solve() and _solve() in its
  api)
- FromAnyState inherits from FromInitState and overrides solve() and _solve()
- init_solve() added to FromAnyState to be called in _solve(), to ensure solve_from() will already have the domain (and for instance the underlying c++ solver) initialized when called from _solve.

Minor modifications to avoid circular imports
- We import Domain for annotations only when type checking
- We use a dedicated logger in parallel_domains.py to avoid havind to import domains

We adapt all solvers to this new api:
- for solvers already implementing solve_from():
  - derive from FromAnyState
  - remove _solve implementation (already in FromAnyState)
- for LazyAstar and LRTAstar:
  - remove attribute self._from_state
  - derive from FromAnyState
  - implement _init_solve() to init self._domain
  - convert `_solve` into `_solve_from`
- for MAHD:
  - derive FromAnyState, but raises an error if chosen solver cannot do it
- for other solvers: nothing changes.

### DIFF
--- a/examples/gym_jsbsim_greedy.py
+++ b/examples/gym_jsbsim_greedy.py
@@ -11,7 +11,7 @@ from gym_jsbsim.catalogs.catalog import Catalog as prp
 from gym_jsbsim.envs.taxi_utils import *
 
 from skdecide import Solver
-from skdecide.builders.solver import DeterministicPolicies, Utilities
+from skdecide.builders.solver import DeterministicPolicies, FromAnyState, Utilities
 from skdecide.hub.domain.gym import DeterministicGymDomain, GymDiscreteActionDomain
 from skdecide.utils import rollout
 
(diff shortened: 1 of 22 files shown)


fteicht (approved):
Thanks @nhuet !

PR #321, opened 2024-04-02
Remove `Solver` intermediate private methods not to be overriden

nhuet:
The convention is the library is that the user has to override/implement
methods whose names are starting with "_".

The methods wihtout the leading "_" are calling the former, generally by
adding autocast boilerplate.

In Solver class, some methods are duplicated without this purpose in
mind, we simplify it here:

- `_get_domain_requirements()` -> `get_domain_requirements()`
- `_check_domain()` -> `check_domain()`, which continue calling
  `_check_domain_additional()` to be overriden
- `_solve()` -> `solve()`, and `_solve_domain()` -> `_solve()`, so that `solve()` autocast
  `domain_factory` and then call `_solve()` to be implemented


fteicht (approved):
Great, thanks @nhuet !

PR #323, opened 2024-04-05
Update dependency in openap to be compatible with scipy 13.0 (last stable release)

nhuet:
Project openap being seemingly dead, we choose as version for openap a tag on a dedicated fork which includes code from openap 1.3 + imports modification to work with latest version of scipy.


fteicht (approved):
Thanks @nhuet !

PR #324, opened 2024-04-10
Allows unified planning actions with simulated effects to be properly hashed

fteicht:
Wrong usage of Unified Planning's grounder prevented the UP bridge in scikit-decide to properly hash grounded actions. This is corrected in this PR.

In order to unit test the correction, the RL unit test solver of the UP bridge now also reasons about UP simulated action effects.


g-poveda (approved):
Look good to me, you could squash commits (the linting one notably)

PR #328, opened 2024-04-18
poll schumann + plots + tests

nestorcarmona:
- Addition of Poll-Schumann aircraft performance model
- Fixes in flight_planning domain
- Addition of unit tests
- flight_planning domain plots updated to include weight profile
- flight_planning notebook updated with OpenAP and Poll-Schumann models


fteicht (approved):
After extensively iterating offline with you @nestorcarmona on this PR,  I think we can now say that you've proposed a very nice addition to the flight planning domain with this PR. Thank you very much for the great work.

PR #330, opened 2024-04-25
Integrate binder environment in development branches instead of a separate dedicated orphan branch

nhuet:
- We move binder definition into a subdirectory "binder/" of dev branches. 
 That way the binder environment will follow the code modifications without having to push on binder orphan branch.
 We still install the scikit-decide last nightly build when building the binder env. The environment corresponding to master branch 
 will be rebuilt each time nightly build is triggered
- We modify the github workflow to stop pushing on binder orphan branch. 
  - in case of push on master, nothing more is needed.
  - In case of nightly build, we trigger the build of master binder env.
  - In case of release, we update the binder subdirectory and push it to the same tag already generated so that the notebooks are 
    modified to install the latest release when launched on colab
  - we also update the actions to make them use node 20 as node 18 is deprecated
  - we refactor to avoid duplicated workflow by creating reusable workflows to build and deploy the doc
- We update the binder links generated to point to use the correct binder environment


nhuet:
I tried the workflows on my own fork:
- master doc is generated there with proper working links: https://nhuet.github.io/scikit-decide/notebooks/#maze-tutorial
- release doc (for fake 0.11.2)  is generated there with proper working links: https://nhuet.github.io/scikit-decide/version/0.11.2/notebooks/#maze-tutorial
- when pushing on master, the binder env is rebuilt (and thus using the last "nightly" build as expected)

fteicht (approved):

PR #343, opened 2024-04-28
Correct moving average epsilons in C++ rollout-based algorithms

fteicht:
C++ rollout-based algorithms could be prematurely stopped when the epsilon moving average computation had not yet reached the moving average window size. This bug is now corrected in this PR.
Also, Labeled-RTDP deactivates stopping condition based on the epsilon moving average and on the rollout budget when labels are used (sine the latter can guarantee convergence when the root note is labeled as solved).


g-poveda (approved):
lgtm

PR #346, opened 2024-04-30
Update local build instructions in the docs

fteicht:
This PR adds hints in the documentation to locally build scikit-decide with poetry.
More specifically, the documentation now indicates the environment variables that must be set prior to locally building and installing the library with poetry.
Not properly setting those environment variables can result in deactivating some options in scikit-decide, e.g. algorithm parallelism.
Those environment variables are set by the building and deployment github actions but they were not mentioned in the documentation for the people who want to build the library on their own.

### DIFF
--- a/docs/contribute.md
+++ b/docs/contribute.md
@@ -46,6 +46,16 @@ brew install cmake
 brew install boost
 brew install libomp
 ```
+You must also set some environment variables to instruct the build process on how to find boost and openMP.
+For instance, on macOS, it is done via:
+```shell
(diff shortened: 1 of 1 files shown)


fteicht:
This commit adds hints in the documentation to locally build scikit-decide with poetry.
More specifically, the documentation now indicates the environment variables that must be set prior to locally building and installing the library with poetry.
Not properly setting those environment variables can result in deactivating some options in scikit-decide, e.g. algorithm parallelism.
Those environment variables are set by the building and deployment github actions but they were not mentioned in the documentation for the people who want to build the library on their own.

The commit also updates the link to poetry's website and installation script which were outdated.

PR #353, opened 2024-05-10
Update DOSolver for last d-o release and add callback functionality

nhuet:
- Move sgs_policies and gphh to do_solve
  - sgs_policies is not a solver itself and is used only by gphh and do_solver
  - do_solver and gphh are both solvers based on discrete-optimization
  - gphh is to be using discrete_optimization's own GPHH

  => we put stuff related to d-o into do_solver subpackage
- Update build_solver()
- Add possibility to plug a callback : this is a function called at the end of each iteration of the solver, taking as arguments the solver and the domain, and returning True if the solver must be stopped.


g-poveda (approved):
Lgtm

PR #354, opened 2024-05-10
Specify sub-dependencies of nbqa pre-commit hooks

nhuet:
This avoids having differences between official, fork and local repos.


g-poveda (approved):
lgtm

PR #356, opened 2024-05-13
Build macos arm64 wheels on a macos arm64 runner

nhuet:
- This is now possible since macos-latest runner is arm64.
- The cross-build seems not working anymore with cibuildwheel last release (2.18)
- We need to install manually conda (to get proper openmp) as it is not installed on last runner images anymore.
- We upgrade the version of Boost library. Older version have a bug with xcode of macos-14: https://stackoverflow.com/questions/77133361/no-template-named-unary-function-in-namespace-std-did-you-mean-unary-fun

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ concurrency:
 
 env:
   BOOST_DIR: 3rdparty/boost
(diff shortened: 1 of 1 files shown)


g-poveda (approved):
thanks, lgtm

PR #358, opened 2024-05-14
Fix load in stable_baselines wrapper (enables retrain)

neo-alex:
This PR fixes a bug in the Stable Baselines 3 wrapper, which prevented additional training after reloading a checkpoint.


g-poveda (approved):
lgtm

PR #359, opened 2024-05-14
Fix callback for DOSolver

nhuet:
- remove domain from callback signature (to be on par with c++-based solvers)
- update dependency in discrete-optimization so that callbacks are indeed included in d-o


g-poveda (approved):
lgtm

PR #371, opened 2024-05-22
Build the doc with python 3.10

fteicht:
Some parts of the documentation were not generated because they rely on dependencies which require python 3.10 (e.g. Unified Planning bindings), whereas the documention was generated with python 3.9.


g-poveda (approved):
Ok, this way unified_planning is included !

PR #373, opened 2024-05-23
Avoid closing domain of full_multisolve

nhuet:
Some classic control environment from gymnasium call pygame.quit() when closing which prevents further rendering if relaunching an environment needed it and thus leading to an error.


g-poveda (approved):
Ok

PR #377, opened 2024-05-23
Update existing enumerations doc and class attributes doc

nhuet:
- attributes: avoid using "-" to make the parser work properly
- enumerations: put relevant doc in __doc__ attribute of each enumeration element

image


g-poveda (approved):
great!

PR #382, opened 2024-05-25
Remove _init_solve

fteicht:
This PR:
- Remove _init_solve() in the solvers which was not consistently used between the FromInitialState solvers (which did not use it to be initialized before the solve()) and FromAnyState solvers (which had to call it to be initialized before the solve()) ;
- Now construct the C++ solvers in their Python interfaces' constructors instead of in _init_solve(), which has the additional benefits of:
   - removing the exception thrown when the methods calling the C++ solvers were not first initialized through a call to the solve() method beforehand ;
   - being able to reuse search objects of the C++ solver from previous calls to the solve(from_state) method instead of reconstructing the C++ solver in the _init_solve() method which was previously called at each start of the solve(from_state) method


fteicht:
Implemented in PR #399

PR #388, opened 2024-05-28
Use API tokens instead of passwords to publish to PyPI

fteicht:


g-poveda (approved):
to be tested but looks ok

PR #404, opened 2024-07-03
Allow the climbing_slope to be a zero

HamdaHmida:
In case of a diversion, there is no need to a climbing phase, so the climbing_slope should not be a None but it can be a zero


HamdaHmida:
I have Updated the branch, now it should pass the tests

PR #405, opened 2024-07-04
Limit numpy dependency to <2.0.0 while other deps adapt themselves

nhuet:
For now,

- ray.rllib
- stablebaselines3
- unified-planning

have issues with the braking changes from numpy 2.0.0. So we temporarily limit the numpy version to make everything still work.


nhuet:
Have been merged via #406 by side effect.

PR #410, opened 2024-07-22
Fix backward and eccodes issues on MacOs and Ubuntu

fteicht:
This PR:
- Correct the way the C++ Backward library is linked to the scikit-decide library so that it can work on MacOS.
- Install libeccodes-dev on Ubuntu in GitHub action tests, which is needed to run the flight planning unit tests


g-poveda (approved):
lgtm

PR #413, opened 2024-07-29
Add repeated state encoding to the UP domain

HamdaHmida:


skdecide/hub/domain/flight_planning/domain.py
```diff
@@ -1007,7 +1007,7 @@ def set_network(
                 )
                 for i in range(nb_vertical_points)
             ]
-        if climbing_slope:
+        if climbing_slope is not None:
```
dbarbier, line 1010: descending_slope should be modified as well

PR #416, opened 2024-09-03
Fix issues with ray-rllib

nhuet:
- starting from 2.34, some renaming occurred, and the properties holding the previous names (e.g. Algorithm.workers) are not working because `Deprecated` decorator does not work well with python `@property`;
- we need ray>=2.9 so that import `from ray.rllib.algorithms import DQN, PPO, SAC` works, to be used in hyperparameter `algo_class` choices definition.


neo-alex (approved):
Thanks, OK to merge!

PR #421, opened 2024-09-16
Remove macos wheel tags update

nhuet:
- This is not working with latest version of cibuildwheel (2.21)
- This is not necessary anymore (no more crossbuilding and cibuildwheel manages already flags correctly)


g-poveda (approved):
lgtm

PR #424, opened 2024-09-27
Correct bug when termination is None in new TerminationOutcome

fteicht:
Creating a TerminationOutcome object with a termination equal to None raised an exception because self.observation is not defined by the TerminationOutcome class


g-poveda (approved):
lgtm

PR #426, opened 2024-09-30
Simplify annotations as we are now python>=3.9

nhuet:
Starting from 3.9,
- using List, Type, Set, ... from typing is deprecated in favor of list, type, set, ... which can now be followed by []
- all classes existing in collections.abc like Callable should also directy be imported from there (from typing import Callable is now deprecated)

So we change the annotations accordingly, removing unnecessary imports from typing.


neo-alex (approved):
Thanks a lot for this update, OK to merge!

PR #427, opened 2024-10-03
Update dependencies scipy and d-o

nhuet:
- scipy: was always installed even though optional because not put in the proper extras
- d-o: 
  -  is now mandatory (to manage hyperparameters)
  - is updated
  - doc and github workflow updated as minizinc is not required anymore except when running wrapped d-o solvers based on minizinc


nhuet:
This PR will work once next release for d-o is done.

g-poveda (approved):
all good to me

PR #430, opened 2024-10-07
Update scheduling notebook to use callback to time-out a solver

nhuet:
In the scheduling notebook, we time-out a lazy A* solver because we know that it cannot tackle the scheduling problem at hand with too many tasks. 
Until now we were using the `signal` library but this code will not work on windows. Therefore we make use of the callback hooks implemented since scikit-decide 1.0.0 to perform the same time-out.


g-poveda (approved):
lgtm

PR #433, opened 2024-10-15
Adapt to d-o 0.5.0

nhuet:
We adapt the scheduling domains and solvers as a full renaming of modules and classes is done in d-o 0.5.0 (to come).


g-poveda (approved):
lgtm

PR #434, opened 2024-10-18
Fix links to js code for deployed release doc

nhuet:
In docs/.vuepress/config.js, the environment variable DOCS_VERSION_PATH is used to prepend the proper subpath to all javascript paths.

The generation of this variable was lost when build-doc.yml and deploy-doc sub-workflows were created to share the code beteween pr merge and release workflows. (A that moment, it was thought to be used only to deploy the doc at the correct path.)


g-poveda (approved):
lgtm

PR #446, opened 2024-12-03
Add possibility to GraphPPO for multi inputs with Dict spaces (including graphs)

nhuet:


nhuet:
now included in #441.

PR #447, opened 2024-12-05
Remove option -s of pytest tests

nhuet:
Log are now far too big, cannot even be loaded in github interface.


neo-alex (approved):
OK to merge (as discussed together with the team).

PR #462, opened 2025-01-20
Stop testing weather interpolation and fix $\{delta}ISA$ to 0.0

nhuet:
The webservice used to download weather is  not completely reliable and makes the tests fail more and more often. So we stop trying to download those data.

When not relying on weather_interpolator, dISA was not defined so we set it to 0., assuming standard atmosphere.


g-poveda (approved):
lgtm thanks

PR #466, opened 2025-01-28
Use GNN without fixed features reduction for graph -> node case

nhuet:
- observations = graph
- actions = node of the graph
  action space= Discrete space, an action is represented by the node index

We also:
- rename `graph_obs_to_thg_data` into `graph_instance_to_thg_data`, and add
  the reverse `thg_data_to_graph_instance`
- move `get_obs_shape` in preprocessing module
- add an utility function to extract torch model parameters values
- add a debug option to policy to store initial model parameters and
  check that model parameters have actually been updated during training
  (see examples/gnn/gnn_graph2node_sb3_jsp.py)
- move gnn examples into a new directory examples/gnn


neo-alex (approved):
Looks great, thank you!

PR #468, opened 2025-02-10
Implement maskable version of Graph2NodePPO (sb3 framework)

nhuet:
- an action is a node of the observation graph
- we use action masking to predict only applicable actions


g-poveda (approved):
lgtm

PR #469, opened 2025-02-14
Put in common graph domains used by tests

nhuet:
Put graph domains in a conftest.py (see
https://docs.pytest.org/en/latest/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files) This "simplifies" test files and avoid code duplication.

We also make similarly a module in "examples/gnn" with graph domains used.

They do not seem polished enough to be in `skdecide.hub.domains` so we have a version for tests in "tests/solvers/python/contest.py" and another one in "examples/gnn/domains.py".


g-poveda (approved):
lgtm

PR #472, opened 2025-02-21
Implement GNN-based PPO graph->node for ray.rllib framework

nhuet:
We are dealing here with domain whose observations are graphs and actions, nodes of these graphs.

The custom model follows what has been done for sb3 framework:
- we predict values as before with GNN feature extraction + reduction to a fixed number of features + MLP
- we predict actions with a single GNN which predict logits for each node without knowing in advance their number.

To use that feature, we need to set the attribute `graph_node_action` to `True`. The doc of RayRLlib has been enriched to explain what is happening in that case.

The main issue here is that we have also to pad rollout buffers entriescorresponding to `action_dist_inputs` (the logits) as the number of available actions is potentially varying from an observation to another.

In case of action masking, a bit more work is necessary:

- Because the length of action mask (=number of nodes) may vary, the custom models used before (`TorchParametricActionsModel`) does not apply properly since it uses an action embedding needing the (max)
number of actions in advance. We thus use a simpler version of model (`TorchMaskedActionsModel`), much like
what is done for sb3 framework: we predict action logits as done when no action masking applies and only at the end, we apply the mask (by adding log(action_mask) to the logits). The main difference is that no last layer is managed by the
custom model with weights changed for non applicable actions, this is managed instead by the GNN.
- Once again the buffers may list action masks of different sizes so we need to apply padding before concatenation.
- Lastly, the first dummy samples used to initialize weights are generated by rllib from the observation space of the `AsRLlibMultiAgentEnv`. So we need to match the default size chosen for the nodes when converting the graph space into a dict space (to be recognized by rllib) in the action_mask space enriching the observation space. This is done in `_create_agent_obs_space_for_rllib()`.

As more stuff on action masking is done here, we create an arborescence similar to what is done for gnn to regroup code related to it:
- the custom models for action masking are now available in:
  - skdecide/hub/solver/ray_rllib/action_masking/models/tf for tensorflow
  - skdecide/hub/solver/ray_rllib/action_masking/models/torch for torch
- the space conversion utilities (such as keys for true obs and action mask) in skdecide/hub/solver/ray_rllib/action_masking/utils/spaces/space_utils.py


fteicht:
@nhuet I made a fix for the windows compilation error in `master`.  Can you please synchronise your branch with the latest `master` before I re-run the github actions of your pull request? Thanks

nhuet:
> @nhuet I made a fix for the windows compilation error in `master`. Can you please synchronise your branch with the latest `master` before I re-run the github actions of your pull request? Thanks

Done

fteicht (approved):

PR #474, opened 2025-03-04
Implement a scikit-decide domain wrapping plado `Task` for PDDL

nhuet:
Plado is an alternative to UP for simulating (P)PDDL problems, even though without its own planners/solvers:
https://github.com/massle/plado/tree/main

It is in particular way faster to list applicable actions in a given state (and does not need to enumerate all actions).


fteicht (approved):
Thanks a lot @nhuet !

PR #480, opened 2025-04-07
Install cmake<4 in github workflow

nhuet:
Some github runners embed now a cmake>=4.0 which is not yet compatible with the cmake files. So we freeze it to the previous version (3.31.6) thanks to a dedicated github action.

See the issue and the workaround here: https://github.com/actions/runner-images/issues/11926#issuecomment-2779500114


fteicht (approved):
Thanks a lot @nhuet !

PR #488, opened 2025-06-12
Add main features of the library to readme and online doc

nhuet:
The features list is automatically extracted from readme and added to the homepage of the doc. (Thanks to tags `<!--features-list-start-->` and `<!--features-list-end-->`, not to be removed)
So we only need to update the features list on the readme from now.


fteicht (approved):
Thanks a lot @nhuet !

PR #503, opened 2025-07-21
Update icaps24 notebook and simplified cost function if no weather data

nhuet:
- We add a comment to explain how to remove weather data when not available.
- We update `cost()` so that having `wind_interpolator=None` is not an issue anymore


g-poveda (approved):
lgtm

PR #556, opened 2026-04-24
Add option --import-mode=importlib to pytest command line

nhuet:
This is considered as good practice:
https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code


nhuet:
Not working with ray tests + custom class defined in tests modules

PR #559, opened 2026-06-05
Update to latest cmake and c++ submodule library versions

fteicht:
* Update SDK submodules: spdlog v1.12→v1.17 (fixes Clang 21 consteval error in bundled fmt), json v3.11→v3.12 (CMake 4 support), pybind11 v2.11→v3.0 (CMake 4 support)
* Update cmake_minimum_required to 3.26.6...3.29 with CMAKE_POLICY_VERSION_MINIMUM 3.5 to run on CMake 4 while keeping FindBoost and allowing old SDK libs
* Use PYBIND11_FINDPYTHON ON + find_package(Python) for pybind11 v3 compatibility
* Update CI cmake version to 4.0.3 on all platforms, add best-effort latest Xcode selection on macOS


cpp/src/hub/solver/aostar/impl/aostar_impl.hh
nhuet, line 1: For the other files i understand the changes. Here it is less ovious what it is doing and why it was necessary, but i guess you already reviewed it.

nhuet: ok the last commit explains it

PR #560, opened 2026-06-18
Replace use of functools.cache/lru_cache by wrapt.lru_cache on methods

nhuet:
It avoids some pitfalls. See more about it here:
- https://wrapt.readthedocs.io/en/master/bundled.html#lru-cache
- https://docs.python.org/3/library/functools.html#functools.lru_cache
- https://docs.python.org/3/faq/programming.html#faq-cache-method-calls
- https://github.com/python/cpython/issues/102618

For classes with such cached methods, we also override `__getstate__()` to let them be pickable according to
https://github.com/GrahamDumpleton/wrapt/issues/343.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ dependencies = [
     "pynng>=0.6.2",
     "pathos>=0.2.7",
     "discrete-optimization>=0.9.0",
+    "wrapt>=2.2.1",
 ]
 
 [project.urls]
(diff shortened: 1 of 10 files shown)


fteicht (changes requested):
The conflict on uv.lock needs to be resolved @nhuet but otherwise, everything's good!

fteicht (approved):

PR #561, opened 2026-06-18
Temporarily upper bound pytest as 9.1.0 not compatible with pytest-cases

nhuet:
See https://github.com/smarie/python-pytest-cases/issues/385


nhuet:
Actions seem to work fine without this PR

PR #567, opened 2026-06-29
Update scheduling domain and integrate new generic scheduling problem from d-o

nhuet:
### Update scheduling domain:
- Take into acount time windows and time lags constraints
  - for applicable actions
  - for sgs policies
- Fix time lags definition: no min or max timelag is not equivalent with
  a 0 or max_horizon offset
- Fix cost computation for scheduling domains: take elapsed time into account
- Fix task duration mixins
  - avoid overriding methods if no changes
  - use progress_from in methods having it in signature
  - fix annotations: int/float parameters with default value are not to be
    annotated with `Optional` (which suggests a None value could ok)
- Fix rcpsp_sk:
  - Do not allow None for mandatory arguments
  - Deduce tasks_ids from mandatory arguments
  - Optional None arguments are initialized as empty list or dict if
    needed
- Fix annotations
- Update timewindows definition: drop unused arguments from constructors
- Fix check_if_action_can_be_started() time_since_start when aciton is
  START


pyproject.toml
```diff
@@ -33,6 +33,9 @@ dependencies = [
     "discrete-optimization>=0.5.0",
 ]
 
+[tool.uv.sources]
+discrete-optimization = { git = "https://github.com/airbus/discrete-optimization" }
```
g-poveda: ok, after the new release, we just need to change this and it will work, thanks

g-poveda (approved):

PR #572, opened 2026-07-27
Convert do scheduling problems to skdecide scheduling domains

nhuet:
- Move rcpsp domains inside scheduling subpackage
- Move do->sk scheduling converter to hub/domain/scheduling
- Create generic scheduling domain mimicking d-o generic scheduling problem
- Update do->sk converter to emcompass case of generic scheduling problems
- Use converter in rcpsp_sk_parser
- Add tests


g-poveda (approved):
lgtm

